Issue #14, opened 2015-02-27
secureConnectionStart should account for non-HTTPS secure transports


toddreifsteck:
See http://www.stevesouders.com/blog/2014/08/21/resource-timing-practical-tips/

toddreifsteck:
Discussed on call. Closing.

hiroshik850223:
Hi, I agree that secureConnectionStart returns undefined, when we don't use HTTPS. Why is this issue close?

plehegar:
Because we agreed it needs to return 0 instead.

PR #15, opened 2015-02-27
Convert HTTPS to secure transport to ensure it is recorded for secure tr...

toddreifsteck:
...ansports with other schemes

Solves https://github.com/w3c/resource-timing/issues/14

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+﻿<!DOCTYPE html>
 <html lang="en">
 <head>
   <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


igrigorik:
Hmm, are you trying to account for things like QUIC? Also, while we're on the subject...

- Why is this attribute "optional" to begin with? Given the push towards HTTPS, it seems like we should make this mandatory. 
- We also need to add language to address: https://code.google.com/p/chromium/issues/detail?id=449224#c4

Meta: not sure what @jainarvind's status is as editor for this spec. Are you proposing to take over? /cc @plehegar

plehegar:
It has been optional because not all implementations were sure to be able to report the proper timing info. I'm not sure if this is still the case. You'd be better to have a separate issue if we're going to consider changing it.

If Todd is interested in helping to edit the draft, that's fine and we could make it "official" at the next call.

igrigorik:
@plehegar sgtm, opened https://github.com/w3c/resource-timing/issues/16 to track mandatory discussion.

toddreifsteck:
Opened https://github.com/w3c/resource-timing/issues/17 to track the secureConnectionStart language for pre-existing connections.

igrigorik:
lgtm to merge.

nicjansma:
s/trasnport/transport/

PR #18, opened 2015-03-12
Add workerStart attribute to measure SW startup

### LINKED ISSUE #11: Add workerStart attribute to measure (Service)Worker startup time
If the app is ServiceWorker-enabled then all fetches are routed to the SW, where they may be fulfilled from local cache, rewritten, forwarded to remote server, etc. However, the SW script may be terminated at any point and may need to be (re)started prior to servicing the request - e.g. navigation request spools up a worker, which is then killed, and a new worker is started up to service a subresource fetch. 

**The application needs to be able to distinguish and measure the worker startup time.** 

---

Attribute definition:

> **workerStart**: The workerStart attribute must return a DOMHighResTimeStamp [HR-TIME] with the time immediately before the user agent starts the Worker required to service the request. If the worker process is already available, then this attribute must return the same value as fetchStart.

For [processing section](http://w3c.github.io/resource-timing/#processing-model) update (add new step before 3.5): 

> If the current browsing context has a registered Service Worker, immediately before a user agent starts the worker process or forwards the request to an existing worker process, record the current time as workerStart.

Does that look reasonable? Anything I'm forgetting? `</handwaving>`

_P.S. Once we iron out the language here, we'll need to make a similar update to Navigation Timing.._

/cc @jakearchibald @slightlyoff @KenjiBaheux

igrigorik:
Closes #11. 

- missing registration results in zero value (i.e. SW not used)
- set to time when worker is run or fetch is fired at worker

@slightlyoff could you sanity check the wording around active workers and matching registrations? Does that look sane? 

@plehegar I've removed the HTTP GET clause in processing because its unnecessarily restrictive (SW sees all fetches regardless of the HTTP verb). Let me know if this is an issue... it seems like a bug in current spec - e.g. it would imply that POST'ed XHRs are not reported within RT.


toddreifsteck:
LGTM

plehegar:
In general, it's a bad idea to have optional features imho. We did make secureConnectionStart optional in nav timing and I don't believe it served anyone. If SW isn't used or supported, it should return 0 imho.

igrigorik:
@plehegar sgtm, removed optional in 9b63c9a.

slightlyoff:
LGTM++

PR #19, opened 2015-04-14
Surface failed fetches in the performance timeline

### LINKED ISSUE #12: Clarify presence of requests that don't return a response
https://lists.w3.org/Archives/Public/public-web-perf/2015Feb/0065.html

> There's some differences in the way browsers treat requests that don't
return a response. FF Nightly and IE 11 both create PerformanceResourceTiming entries for
these requests, whereas Chrome & Canary don't. The RT spec isn't explicit as to whether requests that miss a response should be included. -- @andydavies

igrigorik:
- fetches that are blocked (CSP, CORS, etc) are omitted
- fetches aborted due to network/other errors must be included
- failed fetches must surface initialized attributes up to point of
  failure

Closes #12.

### DIFF
--- a/index.html
+++ b/index.html
@@ -257,12 +257,9 @@ official term ECMAScript, since the term JavaScript is more widely known. [[ECMA
 
 <section id="resources-included">
 <h3>Resources Included in the <code><a href="#performanceresourcetiming">PerformanceResourceTiming</a></code> Interface</h3>
-<p>
-	All resources <a title='fetch' href="http://www.w3.org/TR/html5/infrastructure.html#fetch">fetched</a> by the current <a href="http://www.w3.org/TR/html5/browsers.html#browsing-context">browsing</a> [[!HTML5]] or worker [[!WORKERS]] context's MUST be included as <a href="#performanceresourcetiming">PerformanceResourceTiming</a> objects in the <a href="http://www.w3.org/TR/performance-timeline/#sec-performance-timeline">Performance Timeline</a> of the relevant context. Resources that are
-    retrieved from <a href="http://www.w3.org/TR/html5/browsers.html#relevant-application-cache" title='relevant application cache'>relevant
-application caches</a> or local resources MUST be included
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -257,12 +257,9 @@
...
+
```
yoavweiss: Can we avoid extremely long lines in the future? It makes it so much harder to review changes.

yoavweiss:
LGTM, but interested in @mikewest's opinion as well.

index.html
mikewest: 80 columns FTW!

mikewest:
How do you deal with redirects? That is, do you expose the endpoints to which redirection directs, or do you tie everything to the original url that the UA fetched?

igrigorik:
@mikewest currently, we don't expose redirection end-points (see [step in 19](https://rawgit.com/w3c/resource-timing/0eb0f6997fc3f8a70a556212b45fa9ce5cfe7631/index.html#process) for full details). 

That said, we do have an open discussion to change this, see: https://lists.w3.org/Archives/Public/public-web-perf/2015Feb/0059.html

toddreifsteck:
LGTM

igrigorik:
Re, redirects, see: https://github.com/w3c/resource-timing/issues/21

index.html
```diff
@@ -257,12 +257,9 @@
 
 <section id="resources-included">
```
Safinssssss: king

index.html
Safinssssss: king

eligrey:
Was Chrome ever updated to reflect this spec change?

Screen Shot 2020-05-04 at 12 38 44 PM

npm1:
For 404 at least it was supposed to be fixed, per https://bugs.chromium.org/p/chromium/issues/detail?id=883400. Please comment on that issue if you're not seeing that fixed in a recent Chrome version.

PR #33, opened 2015-09-21
Size attributes: switch to unsigned long, and clarify TAO restrictions

igrigorik:
Addresses feedback in https://bugzilla.mozilla.org/show_bug.cgi?id=1154309#c49: 

- use `unsigned long` for *Size attributes
- clarify that *Size attributes are subject to TAO


bzbarsky:
You want `unsigned long long`, because resources (esp. videos) can in fact be larger than 4GB in size.

I still need to look at the cross-origin bits.

igrigorik:
@bzbarsky good call, fixed.

bzbarsky:
The timing allow check bits look reasonable at first glance.

toddreifsteck:
LGTM. Merging

Issue #21, opened 2015-04-29
RT should handle multi-request fetches


igrigorik:
Discussion: https://lists.w3.org/Archives/Public/public-web-perf/2015Feb/0059.html

igrigorik:
Opened a bug on Fetch: https://github.com/whatwg/fetch/issues/75

noamr:
I think having redirects as their own resource entries is doable, now that FETCH and RT are integrated.

I suggest giving them `redirect` as an `initiatorType`, have them totally not exposed when the overall FETCH is cross-origin and fails TAO (or navigation timing with cross-origin redirects), and add a property such as `originalURL` that points to the first URL of a request.

`preflights` in the context of TAO-enabled cross-origin is a bit tricky, I'm not sure that when TAO is enabled it means that the 3rd party domain is OK with exposing the detail of how long in took to perform the preflight.

noamr:
Note also the discussion in https://github.com/w3c/resource-timing/issues/223.
When we add more "TAO-protected" information to resource timing entries, we retroactively change the meaning of existing uses of `TAO` headers.

I believe that at least as a first step, this information should be only available for same-origin, regardless of TAO, until some versioning scheme or some other solution is found for "TAO yesterday and TAO today mean two different things".

kevinfarrugia:
A use-case I have come across for recording *every* request would be to allow you to access `server-timing` headers from the previous navigation in a redirect chain.

--- Example ---

When navigating to a URL, https://example.com/doSomething:

**Response**
```
HTTP/1.1 302
location: /results
server-timing: foo; dur=100
```

The browser is now on https://example.com/results. `performance.getEntriesByType("navigation")` should allow a means to access the `serverTiming` array for the redirect navigation. Currently this is not possible as only the last request in the sequence is available on NT.

PR #36, opened 2015-09-23
improve initiatorType

### LINKED ISSUE #8: Add Request Method attribute
We at Vimeo are working on uploading straight to Google Cloud Storage. Occasionally I notice OPTIONS requests taking a long time to respond (10-30 sec).
 
Here is one example: https://gist.github.com/vayam/d752cf31eeedc4849766#file-gistfile1-txt-L2164
 
I would like to log OPTIONS request timings so I can report to Google Cloud support. Unfortunately, the HTTP method is not recorded in the resource timing spec. Right now I can't distinguish b/w PUT vs OPTIONS. 

From the spec `The name attribute must return the resolved URL of the requested resource.` 
I propose adding `method` attribute to it.

I would like to hear your feedback on this.

/cc @toddreifsteck

igrigorik:
- 'fetch' for fetch() initiated requests
- 'preflight' for CORS-preflight requests
- 'other' to catch all other cases

closes #8 

/cc @toddreifsteck @annevk

### DIFF
--- a/index.html
+++ b/index.html
@@ -355,6 +355,9 @@ The <a href="#performanceresourcetiming">PerformanceResourceTiming</a> interface
 <p>If the initiator is a CSS resource downloaded by the <code><a href="http://www.w3.org/TR/css-syntax-3/#consume-a-url-token">url()</a></code> syntax [[!CSS-SYNTAX-3]], such as <code>@import url()</code> or <code>background: url()</code>,
 on getting, the <code><a href="#widl-PerformanceResourceTiming-initiatorType">initiatorType</a></code> attribute MUST return the DOMString <code>"css"</code>.</p>
 <p>If the initiator is an XMLHttpRequest object [[!XMLHttpRequest]], on getting, the <code><a href="#widl-PerformanceResourceTiming-initiatorType">initiatorType</a></code> attribute MUST return the DOMString <code>"xmlhttprequest"</code>.</p>
+<p>If the initiator is the <a href="https://fetch.spec.whatwg.org/#dom-global-fetch">Fetch method</a> [[!Fetch]], on getting, the <code><a href="#widl-PerformanceResourceTiming-initiatorType">initiatorType</a></code> attribute MUST return the DOMString <code>"fetch"</code>.</p>
+<p>If the initiator is the <a href="https://fetch.spec.whatwg.org/#cors-preflight-fetch">CORS-preflight fetch</a> [[!Fetch]], on getting, the <code><a href="#widl-PerformanceResourceTiming-initiatorType">initiatorType</a></code> attribute MUST return the DOMString <code>"preflight"</code>.</p>
+<p>Otherwise, on getting, the initiatorType attribute MUST return the DOMString <code>"other"</code>.</p>
 </dd>
 
 <dt>readonly attribute DOMString nextHopProtocol</dt>


annevk:
1. It seems the layering here is wrong. All this timing stuff should be refactored in terms of Fetch.
2. "preflight" seems too generic. We might have HSTS preflights soon, so I'd recommend "cors-preflight".
3. I'm a bit worried about exposing preflights to content. Did that go through security review? What are the limitations for failed fetches and such, at what point are they exposed?

toddreifsteck:
The content within this update LGTM. Please follow up on @annevk's concerns before merging.

igrigorik:
> It seems the layering here is wrong. All this timing stuff should be refactored in terms of Fetch.

Agree, and we should probably move this into a separate issue.. But, since we're on the subject, how do you see RT interacting with Fetch? Would it make sense to slot an extra step somewhere between steps 4-5 under [HTTP fetch](https://fetch.spec.whatwg.org/#http-fetch)? There is some complexity to untangle here with redirects as well -- see https://github.com/w3c/resource-timing/issues/21.

Also, for `initiatorType`, how would we set that? The [Request.initiator](https://fetch.spec.whatwg.org/#concept-request-initiator) doesn't provide what we need here; [Request.type](https://fetch.spec.whatwg.org/#concept-request-type) is a bit closer but also doesn't give us exactly what we need.

> "preflight" seems too generic. We might have HSTS preflights soon, so I'd recommend "cors-preflight".

Hmm, trying to wrap my head around what an HSTS preflight is/does. Any threads or references that I can catch up on?

> I'm a bit worried about exposing preflights to content. Did that go through security review? What are the limitations for failed fetches and such, at what point are they exposed?

This is same discussion as https://github.com/w3c/resource-timing/issues/12.

annevk:
I'm not sure how resource timing is done best. Perhaps "fetch groups" can be reused somehow and we annotate the "fetch records" in such groups with timing data? And then resource timing becomes just the API wrapper around those fetch groups?

I suspect `initiatorType` could be an annotation on a "fetch record". If it requires more, we'd need an extra field on request or maybe extend request's initiator field.

HSTS preflights are discussed here to some extent: https://lists.w3.org/Archives/Public/public-webappsec/2015Aug/thread.html#msg95

igrigorik:
> I'm not sure how resource timing is done best. Perhaps "fetch groups" can be reused somehow and we annotate the "fetch records" in such groups with timing data? And then resource timing becomes just the API wrapper around those fetch groups?

Hmm.. Do redirect fetches, preflights, etc, get their own fetch records? How, if at all, is the navigation fetch surfaced? Also, a lot of the metrics we're talking about here are connection-oriented (DNS, TCP, etc), and don't have clean hooks into Fetch -- a lot of things to untangle here.

That said, I'm guessing some combination of fetch group + fetch observer (https://github.com/whatwg/fetch/issues/65) should do the trick... ~ wait for notification that fetch has finished, create the RT entry and queue it for perf obsevers + relevant buffers.

> HSTS preflights are discussed here to some extent: https://lists.w3.org/Archives/Public/public-webappsec/2015Aug/thread.html#msg95

Interesting. Thanks for the link.

---

Taking a step back here.. Seems like before we can merge this update we really ought to resolve #12 (i.e. whether surfacing cors preflights is something we ought to be doing), and investigate Fetch rewrite. As such, I propose we: close this, open a new bug for Fetch rewrite, and note in #12 that we should add new initiatorType once that's resolved (and if still relevant).

annevk:
Redirects are part of the same overall fetch, since they're resolved somewhat atomically for security reasons. But just as redirects manipulate the url list, they could add information to fetch records quite easily. Would need careful scrutiny though.

I'm okay with splitting this up.

igrigorik:
Closing, let's continue in https://github.com/w3c/resource-timing/issues/39.

PR #37, opened 2015-09-24
Queue the object for performance observers

plehegar:


index.html
```diff
@@ -797,13 +797,13 @@
             </li>
         </ol>
       </li>
-      <li>If the <i>primary buffer</i> is full, discard the <a href="#performanceresourcetiming">PerformanceResourceTiming</a> object,
-      created in step <a href="#step-create-object">3.1</a>. Otherwise, add it to the <i>primary buffer</i>.
+      <li>If the <i>primary buffer</i> is full, discard the <a>PerformanceResourceTiming</a> object,
```
igrigorik: I don't think this is the behavior we want. The PerfEntry object should be _always_ delivered to all relevant perf observers. Then, after its queued, we should invoke the logic to append it or drop it based on the status of the buffer.

igrigorik:
LGTM, thanks @plehegar!

PR #38, opened 2015-09-25
Terminology

plehegar:
Shuffled text to have a proper terminology section.
Added "current document" definition.
Removed reference to HR-TIME and replaced with HR-TIME-2

### DIFF
--- a/index.html
+++ b/index.html
@@ -220,7 +220,9 @@ a new value is assigned to it.</p>
 official term ECMAScript, since the term JavaScript is more widely known. [[ECMA-262]]
 </p>
 
-<p>The term <dfn>resource</dfn> is used to refer to these elements and any other user-initiated fetches throughout this specification.</p>
+<p>The term <dfn>resource</dfn> is used to refer to elements and any other user-initiated fetches throughout this specification.
+For example, a resource could originate from XMLHttpRequest objects [XMLHttpRequest], HTML elements [HTML5] such as iframe, img, script, object, embed, and link with the link type of stylesheet, and SVG elements [SVG11] such as svg.
+</p>
 
 <p>
   The term <dfn>cross-origin</dfn> is used to mean non
--- a/index.html
+++ b/index.html
@@ -797,8 +797,9 @@ href="http://www.w3.org/TR/html5/browsers.html#relevant-application-cache" title
             </li>
         </ol>
       </li>
-      <li>If the <i>primary buffer</i> is full, discard the <a>PerformanceResourceTiming</a> object,
-      created in step <a href="#step-create-object">3.1</a>. Otherwise, <a href='http://w3c.github.io/performance-timeline/#dfn-queue-a-performanceentry'>queue</a> the <a>PerformanceResourceTiming</a> object and add it to the <i>primary buffer</i>.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -206,27 +206,41 @@
...
+<p>
+  The term <dfn>current time</dfn> refers to the number of milliseconds
   since the start of navigation of the document until the current moment in time.
+</p>
+<p class='note'>
   This definition of time is based on the High Resolution Time specification
-  [[HR-TIME]] and is different
+  [[HR-TIME-2]] and is different
   from the definition of time used in the Navigation Timing specification
   [[NAVIGATION-TIMING]],
```
igrigorik: Should we make a note that this is for NT1? Newer drafts of NT2 use same definition.

index.html
```diff
@@ -206,27 +206,41 @@
...
+</p>
 <p>
   Throughout this work, all time values are measured in milliseconds since the start of
-  navigation of the document. For example, the start of navigation of the document
-  occurs at time 0. The term <i>current time</i> refers to the number of milliseconds
+  navigation of the document [[!HR-TIME-2]]. For example, the <a href='https://w3c.github.io/navigation-timing/#widl-PerformanceNavigationTiming-startTime'>start of navigation of the document</a>
+  occurs at time 0.
+</p>
+<p>
+  The term <dfn>current time</dfn> refers to the number of milliseconds
```
igrigorik: Not true for workers. I think you want to define this by linking to "time origin" defined in HR-TIME-2.

index.html
```diff
@@ -206,27 +206,41 @@
...
+<p>A DOM attribute is said to be <dfn>getting</dfn> when its value is being
+retrieved (such as by author script), and is said to be <dfn>setting</dfn> when
 a new value is assigned to it.</p>
 
-<p>The term "JavaScript" is used to refer to ECMA262, rather than the
+<p>The term <dfn>JavaScript</dfn> is used to refer to ECMA262, rather than the
 official term ECMAScript, since the term JavaScript is more widely known. [[ECMA-262]]
 </p>
 
+<p>The term <dfn>resource</dfn> is used to refer to these elements and any other user-initiated fetches throughout this specification.</p>
```
igrigorik: Bit confused by this one. What are "these elements"? User-initiated fetches is also ambiguous.

Should we define it as ~ https://fetch.spec.whatwg.org/#concept-fetch?

index.html
plehegar: yep, I realized that but I'd like to do so in a separate PR if you don't mind. I didn't to change the definitions in this mostly editorial PR.

igrigorik: sgtm.

index.html
plehegar: My goal here was only to make sure all term definitions were in the terminology sections. But it does surface the fact that our definitions aren't perfect indeed.

index.html
plehegar: Indeed. And I should the note in the historical section while I'm at it.

plehegar:
To expand a bit more here, I think some of those definitions should be moved to performance timeline, since they might be useful for other performance entry specs.

igrigorik:
LGTM

PR #41, opened 2015-10-20
clarify {transfer,encodedBody,decodedBody}size

### LINKED ISSUE #40: Clarify what transferSize, encodedBodySize should return in various conditions
Since we also return resource timing entries for cached resources, it should be specified if transferSize should return the same value it had during the first load, or just an estimation (body+headers).

Secondly, since we sometimes get 4XX/5XX responses with content, it needs to be clarified how we handle this case. I assume it would still return non-zero values, even if this is not the body of the resource we requested?

Also, while usually encodedBodySize is the same as the Content-Length response header, in some cases it could differ. I assume we want the actual number of bytes on the wire.

igrigorik:
- spell out cache vs network fetch behavior
- notes for non-2xx responses

closes #40

### DIFF
--- a/index.html
+++ b/index.html
@@ -548,7 +548,7 @@ href="http://www.w3.org/TR/html5/browsers.html#relevant-application-cache" title
 
 <dt>readonly attribute unsigned long long transferSize</dt>
 <dd>
-<p>This attribute MUST return the size, in octets received by the client, consumed by the response header fields and the response <a href="http://httpwg.github.io/specs/rfc7230.html#message.body">payload body</a> [[!RFC7230]]:</p>
+<p>This attribute MUST return the size, in octets received from a [HTTP-network fetch](https://fetch.spec.whatwg.org/#http-network-fetch), consumed by the response header fields and the response <a href="http://httpwg.github.io/specs/rfc7230.html#message.body">payload body</a> [[!RFC7230]]:</p>
 <ul>
   <li>If the resource is retrieved from <a href="http://www.w3.org/TR/html5/browsers.html#relevant-application-cache" title='relevant application cache'>relevant application caches</a> or from local resources, it must return zero.</li>
   <li>If there are HTTP redirects or <a href="http://www.w3.org/TR/html5/infrastructure.html#concept-http-equivalent-codes" title='HTTP response codes equivalence'>equivalent</a> when navigating and if all the redirects or equivalent are from the same <a href="https://tools.ietf.org/html/rfc6454#section-4">origin</a> [[!RFC6454]], this attribute SHOULD include the HTTP overhead of incurred redirects.</li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -548,26 +548,36 @@
...
+<ul>
+  <li>If the resource is retrieved from <a href="http://www.w3.org/TR/html5/browsers.html#relevant-application-cache" title='relevant application cache'>relevant application caches</a> or from local resources, it must return zero.</li>
+  <li>If there are HTTP redirects or <a href="http://www.w3.org/TR/html5/infrastructure.html#concept-http-equivalent-codes" title='HTTP response codes equivalence'>equivalent</a> when navigating and if all the redirects or equivalent are from the same <a href="https://tools.ietf.org/html/rfc6454#section-4">origin</a> [[!RFC6454]], this attribute SHOULD include the HTTP overhead of incurred redirects.</li>
+  <li>If the last non-redirected <a href="http://www.w3.org/TR/html5/infrastructure.html#fetch">fetch</a> of the resource is not the same origin as the <a>current document</a>, <a href="#widl-PerformanceResourceTiming-transferSize">transferSize</a> MUST return zero unless the <a href="#timing-allow-check">timing allow check</a> algorithm passes.</li>
```
yoavweiss: Can we distinguish "not allowed" from 0? I can imagine cases where apps would need to know which is which. (and even more cases where apps would consider one as the other)

index.html
```diff
@@ -548,26 +548,36 @@
...
+<p>This attribute MUST return the size, in octets received by the client, consumed by the response header fields and the response <a href="http://httpwg.github.io/specs/rfc7230.html#message.body">payload body</a> [[!RFC7230]]:</p>
+<ul>
+  <li>If the resource is retrieved from <a href="http://www.w3.org/TR/html5/browsers.html#relevant-application-cache" title='relevant application cache'>relevant application caches</a> or from local resources, it must return zero.</li>
+  <li>If there are HTTP redirects or <a href="http://www.w3.org/TR/html5/infrastructure.html#concept-http-equivalent-codes" title='HTTP response codes equivalence'>equivalent</a> when navigating and if all the redirects or equivalent are from the same <a href="https://tools.ietf.org/html/rfc6454#section-4">origin</a> [[!RFC6454]], this attribute SHOULD include the HTTP overhead of incurred redirects.</li>
```
yoavweiss: We probably also need to define what happens when some of the redirects (but not all) are from same origin, or when all of them have timing allowed, but are from different origins.

index.html
```diff
@@ -548,26 +548,36 @@
...
-<p>If the last non-redirected <a href="http://www.w3.org/TR/html5/infrastructure.html#fetch">fetch</a> of the resource is not the same origin as the <a>current document</a>, <a href="#widl-PerformanceResourceTiming-transferSize">transferSize</a> MUST return zero unless the <a href="#timing-allow-check">timing allow check</a> algorithm passes.</p>
+<p>This attribute MUST return the size, in octets received by the client, consumed by the response header fields and the response <a href="http://httpwg.github.io/specs/rfc7230.html#message.body">payload body</a> [[!RFC7230]]:</p>
```
yoavweiss: It is not clear to me from the definition why `transferSize` doesn't include response bodies fetched from the cache, while `encodedBodySize` does, since both are based on `payloadBody`.

index.html
```diff
@@ -548,26 +548,36 @@
...
+<p>This attribute MUST return the size, in octets received by the client, of the <a href="http://httpwg.github.io/specs/rfc7230.html#message.body">payload body</a> [[!RFC7230]], prior to removing any applied <a href="http://httpwg.github.io/specs/rfc7231.html#data.encoding">content-codings</a> [[!RFC7231]]:</p>
+<ul>
+  <li>If the resource is retrieved from <a href="http://www.w3.org/TR/html5/browsers.html#relevant-application-cache" title='relevant application cache'>relevant application caches</a> or from local resources, it must return the size of the payload body prior to removing any applied content-codings.</li>
+  <li>If the last non-redirected <a href="http://www.w3.org/TR/html5/infrastructure.html#fetch">fetch</a> of the resource is not the same origin as the <a>current document</a>, <a href="#widl-PerformanceResourceTiming-encodedBodySize">encodedBodySize</a> MUST return zero unless the <a href="#timing-allow-check">timing allow check</a> algorithm passes.</li>
```
yoavweiss: Again, a "not allowed" signal can be useful

index.html
igrigorik: "in octets received by the client from a [HTTP-network fetch](https://fetch.spec.whatwg.org/#http-network-fetch)"... and define others in terms of [HTTP-network-or-cache fetch](https://fetch.spec.whatwg.org/#http-network-or-cache-fetch)?

index.html
igrigorik: We can't with the current model.. same applies for timing data. If any one of the requests does not pass the timing-allow check we set values to 0. Once we unbundle requests (https://github.com/w3c/resource-timing/issues/21) we can do better.

index.html
igrigorik: How? Note that we'd make it inconsistent with timing attributes which report 0..

igrigorik:
@yoavweiss @valenting PTAL. Does this resolve issues raised in #40?

valenting:
It does for me. Thanks!

yoavweiss:
LGTM. Did you remove the reference to multiple redirect counting (and the fact that they are counted only if all pass Timing-allowed) on purpose? I thought we settled on keeping it (and possibly improving it in the future to account for partially allowed redirect chains)

igrigorik:
@yoavweiss those checks are still there, just under an `li` for each attribute.

Issue #43, opened 2015-10-25
Timing-allow-check not well-defined for same-origin resources


yoavweiss:
The [timing-allow-check](http://w3c.github.io/resource-timing/#timing-allow-check) is not well defined for same-origin resources, and the way it is used in the definition of `transferSize`, `encodedBodySize` and `decodedBodySize` does not limit it to third-party resources.

We should either improve the definition or limit the check's use. (I believe the former would be better)

PR #44, opened 2015-10-25
timing-allow-check handling of same-origin resources

yoavweiss:
Closes https://github.com/w3c/resource-timing/issues/43

As currently defined and used, it is not clear what should happen when a timing-allow-check is run on a same-origin resource. This PR defines that the check should pass for same-origin resources.


igrigorik:
:+1:, thanks!

PR #45, opened 2015-10-25
Fix up a link towards the Fetch spec

yoavweiss:
One of the links towards the fetch spec was accidentally defined in markdown. Changed to an HTML link :)


plehegar:
LGTM

PR #57, opened 2016-05-31
Added startTime to IDL

### LINKED ISSUE #56: startTime is missing from PerformanceResourceTiming entry IDL

yoavweiss:
Closes #56

### DIFF
--- a/index.html
+++ b/index.html
@@ -385,6 +385,7 @@ interface PerformanceResourceTiming : PerformanceEntry {
     readonly        attribute DOMHighResTimeStamp connectStart;
     readonly        attribute DOMHighResTimeStamp connectEnd;
     readonly        attribute DOMHighResTimeStamp secureConnectionStart;
+    readonly        attribute DOMHighResTimeStamp startTime;
     readonly        attribute DOMHighResTimeStamp requestStart;
     readonly        attribute DOMHighResTimeStamp responseStart;
     readonly        attribute DOMHighResTimeStamp responseEnd;


igrigorik:
Isn't it defined on PerformanceEntry? https://www.w3.org/TR/performance-timeline-2/#dom-performanceentry

plehegar:
Might be a dependency issue. v1 relies on the old performance timeline spec at the moment.
https://rawgit.com/w3c/resource-timing/V1/index.html#performanceresourcetiming

If we keep it that way, we're probably going to create more confusion than it is worth. We should consider linking to Performance Timeline v2 from RT v1 instead and convince the Director it's better that way instead of creating confusion.

yoavweiss:
> Isn't it defined on PerformanceEntry? https://www.w3.org/TR/performance-timeline-2/#dom-performanceentry

It certainly is. My bad :/

> Might be a dependency issue. v1 relies on the old performance timeline spec at the moment.
> https://rawgit.com/w3c/resource-timing/V1/index.html#performanceresourcetiming
> 
> If we keep it that way, we're probably going to create more confusion than it is worth. We should consider linking to Performance Timeline v2 from RT v1 instead and convince the Director it's better that way instead of creating confusion.

Seems like current spec's definition of PerformanceEntry leads to https://www.w3.org/TR/performance-timeline-2/#performanceentry

I think we can close this.

plehegar:
indeed. No action needed.

Issue #25, opened 2015-06-24
requestStart definition differ with navtimg


plehegar:
In navtimg, we have:
[[
responseEnd of type DOMHighResTimeStamp, readonly
    This attribute MUST return a DOMHighResTimeStamp with a time value equal to the time immediately after the user agent receives the last byte of the current document or immediately before the transport connection is closed, whichever comes first. The document here can be received either from the server, relevant application caches or from local resources.
]]

In RT, we have:
[[
responseEnd of type DOMHighResTimeStamp, readonly
    This attribute MUST return the time immediately after the user agent finishes receiving the last byte of the resource from relevant application caches or from local resources.
]]

I presume the navtiming is more accurate (it also matches the responseStart definition).

igrigorik:
@plehegar I'm confused by this one.. The bug says requestStart differs but your snips show responseEnd. Which one is the issue? requestStart looks good to me, and it does seem that we should update responseEnd in RT.

toddreifsteck:
Agree that resource timing responseEnd resource timing should be updated.

igrigorik:
I believe #32 should resolve this issue.

plehegar:
LGTM. Thanks.

PR #67, opened 2016-08-19
Define failure scenario timing resolution for same origin and T…

wesleyhales:
wip #12
Links still need to be added, but I wanted to get feedback on wording first

### DIFF
--- a/index.html
+++ b/index.html
@@ -565,6 +565,7 @@ On getting, the <dfn>responseEnd</dfn> attribute MUST return as follows:
 The time immediately after the user agent receives the last byte of the response or immediately before the transport connection is closed, whichever comes first. The resource here can be received either from <a
 href="https://www.w3.org/TR/html5/browsers.html#relevant-application-cache">relevant application caches</a>, local resources, or from the server.
 </li>
+<li>The time immediately before the user agent aborts the fetch due to a network error.</li>
 </ol>
 <p dfn-for='PerformanceResourceTiming'>
 On getting, the <dfn>transferSize</dfn> attribute MUST return as follows:
@@ -768,7 +769,7 @@ Set <a>resource timing buffer full flag</a> to false.
       <li id="step-collection-start">If the user agent is to reuse the data from another existing or completed <a href="https://www.w3.org/TR/html5/infrastructure.html#fetch">fetch</a> initiated from the <a>current document</a>, abort the
           remaining steps.
       </li>
-      <li>If <a href="https://www.w3.org/TR/html5/infrastructure.html#fetch" title='fetch'>fetching</a> the resource is aborted for any reason, abort the remaining steps.       </li>
+
       <li>If the last non-redirected <a href="https://www.w3.org/TR/html5/infrastructure.html#fetch">fetch</a> of the resource fails the <a>timing allow check</a>, the user agent
       MUST set
       <a for="PerformanceResourceTiming">redirectStart</a>,
@@ -780,7 +781,7 @@ Set <a>resource timing buffer full flag</a> to false.
       <a for="PerformanceResourceTiming">requestStart</a>,
       <a for="PerformanceResourceTiming">responseStart</a> and
 	  <a for="PerformanceResourceTiming">secureConnectionStart</a>
-	  to zero and go to step <a href="#step-response-end">17</a>.
+	  to zero and go to step <a href="#step-response-end">16</a>.
       </li>
         <li>Let <a for="PerformanceResourceTiming">domainLookupStart</a>,
        <a for="PerformanceResourceTiming">domainLookupEnd</a>,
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -565,6 +565,7 @@
 The time immediately after the user agent receives the last byte of the response or immediately before the transport connection is closed, whichever comes first. The resource here can be received either from <a
```
igrigorik: @annevk what happens in Fetch if the server closes connection mid-stream (e.g. before all the Content-Length bytes are sent)? Will that trigger "network error" and reject the response promise?

index.html
```diff
@@ -565,6 +565,7 @@
 The time immediately after the user agent receives the last byte of the response or immediately before the transport connection is closed, whichever comes first. The resource here can be received either from <a
 href="https://www.w3.org/TR/html5/browsers.html#relevant-application-cache">relevant application caches</a>, local resources, or from the server.
 </li>
+<li>The time immediately before the user agent aborts the fetch due to a response failure.</li>
```
igrigorik: I think we can use "due to a [network error](https://fetch.spec.whatwg.org/#concept-network-error)". 

However, now I'm wondering what the "immediately before transport connection closed above" is covering, because that also seems like a network error =/

index.html
```diff
@@ -754,6 +755,11 @@
   <ol class='algorithm'>
       <li id="step-create-object">Create a new <a>PerformanceResourceTiming</a> object and
       set <a href="#entryType-attribute">entryType</a> to the DOMString <code>resource</code>.
+        <ul>
+            <li>
+                If the fetch is aborted for any reason, immediately go to step 18 (queue the record).
```
igrigorik: ~"For the remaining steps, if the fetch is...." - I think we need to clarify that this applies for all following steps.

... aborted due to a _network error_ (same ref as above)

igrigorik: Actually, I think this should go into step 7. 

"If fetching the resource is aborted for any reason, abort the remaining steps." > If fetch is aborted due to a network error, go to... queue record.

index.html
```diff
@@ -768,7 +774,7 @@
       <li id="step-collection-start">If the user agent is to reuse the data from another existing or completed <a href="https://www.w3.org/TR/html5/infrastructure.html#fetch">fetch</a> initiated from the <a>current document</a>, abort the
           remaining steps.
       </li>
-      <li>If <a href="https://www.w3.org/TR/html5/infrastructure.html#fetch" title='fetch'>fetching</a> the resource is aborted for any reason, abort the remaining steps.       </li>
+            <li>If <a href="https://www.w3.org/TR/html5/infrastructure.html#fetch" title='fetch'>fetching</a> a cross-domain resource without Timing Allow Origin enabled is aborted for any reason, abort the remaining steps.</li>
```
igrigorik: See note above.

index.html
```diff
@@ -796,11 +802,11 @@
        starts the domain name lookup, record the time as <a for="PerformanceResourceTiming">domainLookupStart</a>.</li>
       <li>Record the time as <a for="PerformanceResourceTiming">domainLookupEnd</a> immediately after the
        domain name lookup is successfully done. A user agent may need multiple retries before that. If
-       the domain lookup fails, abort the remaining steps. </li>
+                the domain lookup fails, record the time immediately before the user agent aborts the same domain or Timing Allow Origin enabled request for resources, otherwise abort the remaining steps.</li>
```
igrigorik: Perhaps... 

"Record the time as _domainLookupEnd_ immediately after the domain name lookup is successfully done, or if the lookup fails and resource passes the _timing allow check_. A user agent may need multiple retries before that. If the domain lookup fails, go to step X."

igrigorik:
👍 ... I think it's on the right track.

index.html
annevk: Not currently. And synthetic responses / content codings make this header unreliable.

If user agents were to implement such logic we might add it, but it seems that for something that streams for a long time, it would be a rather disastrous user experience.

annevk:
(I guess actually adding the logic into Fetch is the next step?)

index.html
igrigorik: Makes sense. @plehegar do you, by any chance, remember why the "or immediately before the transport connection is closed" was added here / was meant to capture? Was this a clause for HTTP/1.0 and lack of keep-alive?

plehegar: This was introduced following https://github.com/w3c/resource-timing/issues/25 , ie to sync with nav-timing. And this was in nav-timing since October 2010 ( https://dvcs.w3.org/hg/webperf/rev/3bb1ac67bfb2 ), based on the f2f ( https://lists.w3.org/Archives/Public/public-web-perf/2010Oct/0016.html ). It says "Add support for scenario where connection is closed."

igrigorik: Interesting thanks Philippe! I'll go on the assumption that it was lack-of-keepalive related.

index.html
```diff
@@ -802,7 +797,7 @@
        starts the domain name lookup, record the time as <a for="PerformanceResourceTiming">domainLookupStart</a>.</li>
       <li>Record the time as <a for="PerformanceResourceTiming">domainLookupEnd</a> immediately after the
        domain name lookup is successfully done. A user agent may need multiple retries before that. If
-                the domain lookup fails, record the time immediately before the user agent aborts the same domain or Timing Allow Origin enabled request for resources, otherwise abort the remaining steps.</li>
+       the domain lookup fails, record the time as domainLookupEnd immediately after the domain name lookup is successfully done, or if the lookup fails and resource passes the <a>timing allow check</a>. A user agent may need multiple retries before that. If the domain lookup fails, set to zero and <a href="#step-final-queue">queue the record</a>.</li>
```
wesleyhales: @igrigorik How do we handle recording the duration for the final condition? Since Step 17 does not happen, should we set duration here or am I missing something?

igrigorik: "Set to zero and go to step 18" instead of calling queue directly? 

Sorry about the confusion, reviewing the [preview](https://cdn.rawgit.com/wesleyhales/resource-timing/0cbce90b377f5ab8e3c2780000b8dc425e7c7cf2/index.html) once more: 
 - Current steps 1-10 are setup.
 - 11-18 is where all the timestamps are recorded.
  - 12: Record the time as domainLookupEnd immediately after the domain name lookup is successfully done. A user agent may need multiple retries before that. If the domain name lookup fails and resource passes the timing allow check, record the time as domainLookupEnd and go to step 18. 
  - 13-17... ~similar as above
  - 18: If responseEnd end is not set, set it to current time. Record the difference between responseEnd and startTime in duration.
  - ... rest as usual.

With above, I think we can also drop step 7 (in which case offset the step numbers above).

WDYT, does that look more reasonable?

igrigorik:
> (I guess actually adding the logic into Fetch is the next step?)

This change is targeting L2, which doesn't integrate with Fetch. In L3 we'll refactor this to hook up with Fetch.

index.html
```diff
@@ -823,7 +823,7 @@
       <li id="step-response-end">Record the time as <a for="PerformanceResourceTiming">responseEnd</a>
       immediately after receiving the last byte of the response.
       <ol>
-        <li>Return to step <a href="#step-connect-start">13</a> if the user agent fails to send the request or receive the entire response, and needs to reopen the connection.
+        <li>Return to step <a href="#step-connect-start">12</a> if the user agent fails to send the request or receive the entire response, and needs to reopen the connection.
```
wesleyhales: What if the UA does not try to reopen the connection... e.g. closes it or times out on the response? Should we say something like "...if the user agent does not attempt to reopen the connection proceed to step 17?"

index.html
wesleyhales: Sounds good to me 👍

index.html
igrigorik: I think current wording is fine. If it doesn't need/want to reopen it'll just continue and set responseEnd on 17?

igrigorik:
Overall, I think this gets us what we ~want.. A few observations:
- we don't have secureConnectionEnd, which means we don't have a clear way to signal end of unsuccessful TLS handshake. I don't think this is something we can or should address in L2 though.
- after re-reading the processing model, we can (read should) do some refactoring there.. But, if we're doing that, we might as well refactor on top of fetch, which we already slated for L3.

@toddreifsteck can you take a pass? 

_(p.s. I'm wondering if we should move this whole thing to L3 and pair it with fetch refactor)_

toddreifsteck:
This update looks good to me. I'm ok with taking the change now or waiting till we refactor.

wesleyhales:
@igrigorik what's the next step from here? Should I squash my commits or will you handle on the merge?

igrigorik:
@wesleyhales apologies about the delay, been offline for the past few weeks. Yes, please squash.

Re, L2/L3: I propose we create an L3 branch and merge this into L3. We can then tackle the Fetch rewrite in a later commit. The reason for L3 is that this new behavior would not pass on any existing L2 implementations, and I think we want to stabilize and ship L2 CR sooner rather than later; we can shift new features into L3.

wesleyhales:
Squashed. Should I submit a new PR to an L3 branch? Shall I go ahead and create it?

igrigorik:
@plehegar my preference would be to keep latest draft on master branch? Should we create an L2 branch and iterate there, and merge this into master / treat it as L3? WDYT?

Issue #73, opened 2016-11-08
Clarify nextHopProtocol value for cached fetch


igrigorik:
https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/Hs5-wPpCPZc/7BVwLbfMAAAJ

> One thing that's a bit unclear about this spec is how it would handle cached loads. Would it return the nextHop for the original fetch or would it return a value like "cache". FWIW at FB we're already differentiating cache via transfer_size = 0. -- @bmaurer 

Inspecting FF's (shipped in M45) implementation: they return the value of the original fetch, which is a sensible implementation given that current definition constrains the returned value to ALPN id's. 

That said, from a developer perspective, returning a value like "cache" or "local" seems (more, I think) sensible. Doing so would remove the need for attribute gymnastics around transferSize (see https://github.com/w3c/resource-timing/issues/42#issuecomment-206508028) to figure out if a resource is cached.. Alternatively, we can return an empty string to indicate that there is no "next hop".

@valenting @toddreifsteck any thoughts or opinions on this one?

toddreifsteck:
I generally prefer 'cache' to empty string as I believe it explicitly describes the "hop" that is taken to get the data.

Curious what FF thinks!

bmaurer:
I'd highly support having something like cache here. It'd be great if browsers had some latitude to expose information about the nature of the cache. Eg cache+memory.

valenting:
>  One thing that's a bit unclear about this spec is how it would handle cached loads. Would it return the nextHop for the original fetch or would it return a value like "cache". FWIW at FB we're already differentiating cache via transfer_size = 0.

This wouldn't always work, as transferSize isn't 0 for revalidated resources. In this case, the value of nextHopProtocol might still be useful indicating protocol used to do a conditional request.

> I'd highly support having something like cache here. It'd be great if browsers had some latitude to expose information about the nature of the cache. Eg cache+memory.

cache+memory, cache+disk, cache+net revalidation are things that might be quite useful to expose. But depending on how much information we want to return, it could be better as a separate attribute.

CC @bzbarsky

bzbarsky:
Is there a specific question I'm being asked here?

valenting:
Maybe you have an opinion on what nextHopProtocol ought to return for cached resources.

bzbarsky:
I don't, sorry.

bmaurer:
[...]

PR #78, opened 2016-11-11
nextHopProtocol is an empty string for local fetch

igrigorik:
Processing algorithm is already aligned with above:
- Step 2 sets nextHopProtocol to empty string.
- Step 10 jumps to step 15 and over step 13, which sets the attribute
  value if the resource is served from local cache.

Closes https://github.com/w3c/resource-timing/issues/73.

/cc @valenting @toddreifsteck @bmaurer


valenting:
👍

igrigorik:
@toddreifsteck if this looks good, please feel free to merge! :)

yoavweiss:
LGTM

toddreifsteck:
Do we have a crisp issue tracking that cache is not directly exposed? I'm hesitant to agree to this unless we close on the direction for that discussion first. Was the thinking to add another property? Can we open an issue for that please?

igrigorik:
@toddreifsteck I believe the cache discussion is orthogonal and non-blocking to this. If there is interest, we can explore how and if we can expose more "what type of cache the resource came from", but that's an L3 + Fetch integration discussion. The question to address here is what nextHopProtocol should report when there is no network hop... which, this PR proposes set to the empty string.

toddreifsteck:
Discussed on 11/17 Web Perf Working Group call. Conclusion was that nextHopProtocol is only about the ALPN id and should not have additional meanings added. Mergning

Issue #55, opened 2016-05-24
Is it well defined when the "add a PerformanceResourceTiming entry" algorithm should run?


yoavweiss:
I couldn't find when the "add a PerformanceResourceTiming entry" algorithm is defined to run, and I have ran into trouble in WebKit with an example similar to the one in Example 2 (only with XHR), where at the onload event, there is expectation that the entry should already be there.

Is the algorithm's running point well defined? Do we have tests for that?

igrigorik:
Following the [processing model](https://w3c.github.io/resource-timing/#process): 

- For each resource fetched by the current browsing context, ..., perform the following steps:
 - Create a new PerformanceResourceTiming object and set entryType to the DOMString resource.
 - ... initialize all the timestamps ...
 - After recording duration (once we know responseEnd)
   - *Queue* the PerformanceResourceTiming object.
    - *Add* the PerformanceResourceTiming object.

So, conceptually it's supposed to be added when the response becomes available - e.g. ~same time as XHR success callback. We don't have good hooks for this though in L2.. At least, not until we integrate with fetch. That said...

> I have ran into trouble in WebKit with an example similar to the one in Example 2 (only with XHR), where at the onload event, there is expectation that the entry should already be there.

I think that's a reasonable read based on current definition of "add a perf entry", but perhaps something we should discourage: synchronous API's are an anti-pattern and developers shouldn't rely on Perf Timeline exposing new entries in synchronous fashion (e.g. XHR success != perf entry is immediately available). In fact, we already do this for PerformanceObserver: https://w3c.github.io/performance-timeline/#dfn-queue-a-performanceentry

yoavweiss:
Based on my reading of this, adding the entry either before or after the onload is permitted and creates an observable implementation difference.

Your "synchronous APIs are an anti-pattern" argument makes sense to me, and I agree we should discourage it. As such, can we define that the entry is added after the onload event?

In any case, it'd probably be a good idea to modify the example so that it doesn't rely on that currently-implementation-specific behavior.

igrigorik:
> Your "synchronous APIs are an anti-pattern" argument makes sense to me, and I agree we should discourage it. As such, can we define that the entry is added after the onload event?

[...]

PR #79, opened 2016-11-15
Recommendation for when add/queue steps should run

igrigorik:
Closes https://github.com/w3c/resource-timing/issues/55.

@yoavweiss @toddreifsteck @plehegar ptal.

### DIFF
--- a/index.html
+++ b/index.html
@@ -855,9 +855,16 @@ Set <a>resource timing buffer full flag</a> to false.
       </li>
       <li><a href='https://www.w3.org/TR/performance-timeline-2/#dfn-queue-a-performanceentry'>Queue</a> the <a>PerformanceResourceTiming</a> object.
       </li>
-      <li><a data-lt='add a PerformanceResourceTiming entry'>Add</a> the <a>PerformanceResourceTiming</a> object.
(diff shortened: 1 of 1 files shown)


toddreifsteck:
Not a fan. Please refer to #55 comments. Happy to discuss on 11/17 call.

igrigorik:
(Rough) summary from our call last week...

- Ideally, we should have a MUST: that allows us to write clean interop tests, avoids surprises for developers, and so on.
- Currently, the spec leaves this behavior undefined and browser implementations differ + sometimes vary based on type pf resource. That is, sometimes you _can_ count on the entry being queued before load event is fired, but sometimes that's also not the case in the same browser; different browsers have different permutations of this behavior. As such, in practice, what you _can_ count on is that the entry will be available _sometime_ after the resource has loaded + been processed.
- Moving forward, we are encouraging developers to use PerformanceObserver, which actually has very similar semantics as above: it __does not__ make any guarantees about when the even is delivered, and the processing algorithm specifies that such events _should be_ queued from a low priority queue to minimize contention with other (more critical) browser work.
- If we change the behavior to a MUST and current implementations update.. Will they break existing code that relies on events being available at a particular point in time? This is very plausible, but there is high likelihood this code already doesn't work in some browsers, and it will not work in any new implementations after the spec update is made -- e.g. Safari, if/when they ship RT2.

As next steps, we identified...
- @toddreifsteck to check with his team on queueing after ~load event
- ...?

@yoavweiss @toddreifsteck what other questions do we need to answer to resolve this?

igrigorik:
As per discussion on the call today.. Opened #82 and updated this PR to link to it. Merging.

Issue #69, opened 2016-10-10
Initiator is not well defined


yoavweiss:
I've been looking at specific ways to add more detailed initiator info that would enable getting the dependency tree info from resource-timing (as briefly discussed during TPAC).

When digging deeper into the spec and the current definitions, I found some unfortunate surprises:
- The term "initiator" is not really defined in the spec
- `initiatorType` sometime points at the type of the resource itself, and sometimes points at the type of the initiator (specifically, with the "css" initiator type)
- Fetch contains the concept of "[initiator](https://fetch.spec.whatwg.org/#concept-request-initiator)", which is not related to what we need here in any way.

Furthermore, for the purposes of a dependency tree, initiator definition needs to vary from the one practiced with `initiatorType` (and match more closely what happens in the various browser dev tools):
- If an element triggered a resource download and was part of the HTML's markup, the "dependency initiator" should be the HTML itself
- If an element triggered a resource download, but was added to the HTML by a JS file, the "dependency initiator" should be the JS resource (or maybe JS resources, if the addition was the result of multiple JS resources executing)
- Cases which now get "css" as initiator type, the "dependency initiator" should point to the CSS resource.
- If a download was triggered by XHR/Fetch, the "dependency initiator" should point to the JS resource(s).

I think that for the purposes of `initiatorType`, we'd better define the behavior as it is currently shipped.

For the purposes of the "dependency initiator", we probably need to define something slightly different. As `initiatorType` is already shipped and the term "initiator" seems to be overloaded, we probably also want to find a new name.

igrigorik:
> I think that for the purposes of initiatorType, we'd better define the behavior as it is currently shipped

Agreed, I don't think we can change it significantly. That said, what would you propose we change or update with respect to current definition?

Aside: I propose we focus this thread on the above and resolve it separately from a design discussion on what we could/should do to tackle the "dependency tree" use case. For the latter, let's open a new thread and describe the use case(s) and what's needed to enable them.

igrigorik:
Taking a second pass at this... 

> I think that for the purposes of initiatorType, we'd better define the behavior as it is currently shipped.

Spec: https://www.w3.org/TR/resource-timing-2/#dom-performanceresourcetiming-initiatortype

> On getting, the initiatorType attribute must return one of the following DOMString:
> 
> - The same value as the localName of that element [DOM], if the initiator is an element.
[...]

PR #80, opened 2016-11-17
Clarify initiatorType

igrigorik:
Based on the discussion earlier today on the webperf call...

- Added: `fetch`, `beacon`, `navigation` as valid values. The last case is
  important for ServiceWorker cases; Nav Timing 2 will report same
  value.
- If none of the specified cases match, the default fallback is `other`.

Closes https://github.com/w3c/resource-timing/issues/69.

### DIFF
--- a/index.html
+++ b/index.html
@@ -408,12 +408,12 @@ On getting, the <dfn>initiatorType</dfn> attribute MUST return one of the follow
 </p>
 <ul>
 <li>The same value as the <code><a href="https://www.w3.org/TR/dom/#concept-element-local-name">localName</a></code> of that
-<code><a href="https://www.w3.org/TR/dom/#concept-element">element</a></code> [[!DOM]], if the initiator is an <code><a href="https://www.w3.org/TR/dom/#concept-element">element</a></code>;</li>
-<li><dfn id='css-resource'><code>"css"</code></dfn>, if the initiator is a CSS resource downloaded by the <code><a href="https://www.w3.org/TR/css-syntax-3/#consume-a-url-token">url()</a></code> syntax [[!CSS-SYNTAX-3]], such as <code>@import url()</code> or <code>background: url()</code>;</li>
+<code><a href="https://www.w3.org/TR/dom/#concept-element">element</a></code> [[!DOM]], if the request is a result of processing the <code><a href="https://www.w3.org/TR/dom/#concept-element">element</a></code>;</li>
+<li><dfn id='css-resource'><code>"css"</code></dfn>, if the request is a result of processing a CSS <code><a href="https://www.w3.org/TR/css-syntax-3/#consume-a-url-token">url()</a></code> directive [[!CSS-SYNTAX-3]], such as <code>@import url()</code> or <code>background: url()</code>;</li>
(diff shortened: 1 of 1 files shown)


igrigorik:
@toddreifsteck @yoavweiss @plehegar ptal.

index.html
```diff
@@ -404,13 +404,17 @@
...
-On getting, the <dfn>initiatorType</dfn> attribute MUST return one of the following <code>DOMString</code>:
+On getting, the <dfn>initiatorType</dfn> attribute MUST return one of the following <code>DOMString</code> values:
 </p>
 <ul>
 <li>The same value as the <code><a href="https://www.w3.org/TR/dom/#concept-element-local-name">localName</a></code> of that
-<code><a href="https://www.w3.org/TR/dom/#concept-element">element</a></code> [[!DOM]], if the initiator is an <code><a href="https://www.w3.org/TR/dom/#concept-element">element</a></code>.</li>
-<li><dfn id='css-resource'><code>"css"</code></dfn>, if the initiator is a CSS resource downloaded by the <code><a href="https://www.w3.org/TR/css-syntax-3/#consume-a-url-token">url()</a></code> syntax [[!CSS-SYNTAX-3]], such as <code>@import url()</code> or <code>background: url()</code>.</li>
-<li><dfn id='xhr-resource'><code>"xmlhttprequest"</code></dfn>, if the initiator is an XMLHttpRequest object [[!XMLHttpRequest]].</li>
+<code><a href="https://www.w3.org/TR/dom/#concept-element">element</a></code> [[!DOM]], if the initiator is an <code><a href="https://www.w3.org/TR/dom/#concept-element">element</a></code>;</li>
```
yoavweiss: This still defines `initiatorType` according to "initiator", which is not defined anywhere. I don't have easy solutions to defining it here tho (there are no equivalent Fetch hooks, other than creating a map of Fetch `type` and `destination` to `initiatorType`, but even with that, some types fall off the edges, e.g. "beacon")

index.html
```diff
@@ -404,13 +404,17 @@
...
+<li><dfn id='css-resource'><code>"css"</code></dfn>, if the initiator is a CSS resource downloaded by the <code><a href="https://www.w3.org/TR/css-syntax-3/#consume-a-url-token">url()</a></code> syntax [[!CSS-SYNTAX-3]], such as <code>@import url()</code> or <code>background: url()</code>;</li>
+<li><dfn id='navigation-resource'><code>"navigation"</code></dfn>, if the request is a <a href="https://fetch.spec.whatwg.org/#navigation-request">navigation request</a>;</li>
```
yoavweiss: Is this defined for navigation timing's sake?

igrigorik: There's also ServiceWorker.. which intercepts navigation requests via RT and needs to be able to distinguish them as such.

index.html
igrigorik: s/initiator/if the request is a result of processing the element.

or, some such?

yoavweiss: Definitely better than current definition

igrigorik:
@yoavweiss updated, ptal.

plehegar:
LGTM

yoavweiss:
LGTM2

toddreifsteck:
@nicjansma Can you confirm that this matches your data?

nicjansma:
@toddreifsteck Looks great

Issue #63, opened 2016-07-22
Clarify definition of responseStart


igrigorik:
From https://bugs.chromium.org/p/chromium/issues/detail?id=568024: 

http://w3c.github.io/resource-timing/#dom-performanceresourcetiming-responsestart

> Is the first byte the first byte of the first challenge, or the first byte of the actual headers?  What about 100 continue responses?  Do those really count, or do we care about the actual headers for the response? 

I'm inclined to say that the challenge bytes / 100 Continue are valid "responseStart" bytes - i.e. responseStart should capture the time of receipt of 100 Continue / challenge.

@toddreifsteck @annevk how do you guys handle this in your implementations?

> if H2 reads only a partial frame.. 

Is that a valid responseStart marker, or should it reflect receipt of a full frame? Personally I'm leaning towards the latter.

annevk:
We should figure out Fetch integration. This would fall out of that. Not sure about what Fx does.

toddreifsteck:
Edge/IE: For HTTP/2, it is the first full header. For others, it is the first completed receive of bytes. Which could range in size depending on the underlying network.

igrigorik:
@toddreifsteck to clarify, "first full header" would be the first full HEADERS frame.. regardless of whether that frame is carrying `:status:` 100, 200, etc?

toddreifsteck:
Yes, you are correct. The first full HEADERS frame.

igrigorik:
Reviewing the code... 

- From what I can see, Fx sets responseStart as soon as bytes are received: https://github.com/mozilla/gecko-dev/blob/553a2da922bb1d1134ede1f31e609e0d5f6f6a5e/netwerk/protocol/http/nsHttpTransaction.cpp#L758. @mcmanus what happens for the H2 case?
- Chrome's current behavior is to report time when we finished receiving header bytes on the client: https://cs.chromium.org/chromium/src/third_party/WebKit/Source/core/timing/PerformanceTiming.cpp?sq=package:chromium&dr=C&l=232-244 - note the FIXME.

@annevk re, fetch integration: it looks like we'd want to extend [5.6 step 4](https://fetch.spec.whatwg.org/#http-network-fetch) to provide a hook for when first header bytes are received?

annevk:
Yeah, seems like it. Does this get reset when there's a redirect then or HTTP authentication?

[...]

PR #83, opened 2016-11-30
clarify responseStart

igrigorik:
Discussion:
- https://github.com/w3c/resource-timing/issues/63
- https://www.w3.org/2016/11/30-webperf-minutes.html

Closes https://github.com/w3c/resource-timing/issues/63.

### DIFF
--- a/index.html
+++ b/index.html
@@ -562,11 +562,15 @@ On getting, the <dfn>responseStart</dfn> attribute MUST return as follows:
 </p>
 <ol class="algorithm">
 <li>
-The time immediately after the user agent receives the first byte of the response from <a
+The time immediately after the user agent's HTTP parser receives the first byte of the response (e.g. frame header bytes for HTTP/2, or response status line for HTTP/1.x) from <a
 href="https://www.w3.org/TR/html5/browsers.html#relevant-application-cache">relevant application caches</a>, or from local resources or from the server if the last non-redirected <a href="https://www.w3.org/TR/html5/infrastructure.html#fetch">fetch</a> of the resource passes the <a>timing allow check</a> algorithm.
 </li>
 <li>zero, otherwise.</li>
 </ol>
+<aside class="note">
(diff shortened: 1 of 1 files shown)


igrigorik:
@toddreifsteck @yoavweiss ptal. Hopefully that captures what we discussed on the call today.

index.html
```diff
@@ -562,11 +562,15 @@
...
+The time immediately after the user agent's HTTP parser receives the first byte of the response (e.g. frame header bytes for HTTP/2, or response status line for HTTP/1.x) from <a
 href="https://www.w3.org/TR/html5/browsers.html#relevant-application-cache">relevant application caches</a>, or from local resources or from the server if the last non-redirected <a href="https://www.w3.org/TR/html5/infrastructure.html#fetch">fetch</a> of the resource passes the <a>timing allow check</a> algorithm.
 </li>
 <li>zero, otherwise.</li>
 </ol>
+<aside class="note">
+  <p>There are many reasonable definitions of "receipt of first byte": as seen by the kernel, by the browsers network stack / HTTP parser, by the renderer, and so on. The above definition is with respect to the HTTP parser, and thus may include queuing time inside of client's kernel buffers and other steps that may preceed receipt of data by the HTTP parser.</p>
+  <p>For fetches composed of multiple requests (e.g. preflights, authentication challenge-response, redirects, and so on), the reported <code>responseStart</code> value is that of the last request.</p>
```
yoavweiss: What is the value in cases like "100 continue" (or "early hints" in the future)? That's a response to the last request, but not the last response... 
By current definition their arrival will mark `responseStart`. I just want to make sure that this is as intended (and maybe we want to add a note to that effect as well).

igrigorik:
@yoavweiss good catch, updated. How does that look now?

yoavweiss:
LGTM

igrigorik:
@yoavweiss thanks!
@toddreifsteck ptal, and hit merge if this seems reasonable :)

toddreifsteck:
This looks good. Do we need tests?

PR #97, opened 2017-02-10
Merge V2 into gh-pages

yoavweiss:
It seems we have committed many bug fixes to the V2 branch without merging them into the gh-pages branch, which is what is published as an editor draft.

That have already resulted in implementor confusion, where resolved issues were raised again :/

This PR merges those V2 changes back to the editor's draft.

In the future we should make sure that all new work happens on gh-pages, and then cherry-picked into the relevant version branch, rather than the other way around.


igrigorik:
For background and context: https://lists.w3.org/Archives/Public/public-web-perf/2017Feb/0007.html

index.html
```diff
@@ -292,7 +293,7 @@
...
...etched</a> by the current <a href="https://www.w3.org/TR/html5/browsers.html#browsing-context">browsing</a> [[!HTML5]] or worker [[!WORKERS]] context's MUST be included as <a>PerformanceResourceTiming</a> objects in the <a href="https://www.w3.org/TR/performance-timeline-2/#sec-performance-timeline">Performance Timeline</a> of the relevant context. Resources that are retrieved from <a href="https://www.w3.org/TR/html5/browsers.html#relevant-application-cache">relevant application caches</a> or local resources MUST be included as <a>PerformanceResourceTiming</a> objects in the <a href="https://www.w3.org/TR/performance-timeline-2/#sec-performance-timeline">Performance Timeline</a> [[!PERFORMANCE-TIMELINE-2]]. Resources for which the <a title='fetch' href="https://www.w3.org/TR/html5/infrastructure.html#fetch">fetch</a> was initiated, but was later aborted (e.g. due to a network error) MUST be included as <a>PerformanceResourceTiming</a> objects in the <a href="https://www.w3.org/TR/performance-timeline-2/#sec-performance-timeline">Performance Timeline</a> and MUST contain initialized attribute values for processed substeps of the <a href="#processing-model">processing model</a>.</p>
```
igrigorik: Note that this will undo some of the L3-specific work we merged earlier, which updated the requirements from ~"UA MAY include aborted fetches" to "MUST".

Stepping back, I think that may actually be the only major difference between current L2 and L3 specs. Also, given that we're planning to rewrite L3 in terms of Fetch.. the L3 update is something we'll have to do redo anyway -- originally we thought we'd merge this into L2, but at TPAC we decided to punt this requirement to L3. 

Long story short, merging this will undo some of the L3 work, but I think that may be OK.. In the meantime, it'll make our life a lot easier to keep v2 and gh-pages in sync. Once we're ready to start working on L3 in earnest, we'll sort out the MAY/MUST bits. 

@toddreifsteck wdyt?

PR #105, opened 2017-03-22
Replaced TAO section with L3 version

plehegar:


plehegar:
@igrigorik @toddreifsteck once this gets merged (you can safely ignore the check failures btw), I'll do the w3c incantations to get it published.

PR #109, opened 2017-04-13
ABNF uses list extension

### LINKED ISSUE #99: Custom ABNF Syntax specification
Introduced in #62 the [spec uses new ABNF syntax](http://yoavweiss.github.io/resource-timing/#timing-allow-origin), while linking to [RFC5234](https://tools.ietf.org/html/rfc5234):

>       Timing-Allow-Origin = 1#( origin-or-null / wildcard )

The `#` syntax is not listed anywhere in the linked RFC5234. It seems to actually be custom w3 extension, described at [https://www.w3.org/Notation.html](https://www.w3.org/Notation.html) but not linked anywhere by any specification that I could find. Please include add some kind of link.

plehegar:
This should fix #99

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/resource-timing/abnf.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/resource-timing/177e382...88cf7ca.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -701,7 +701,7 @@ Set <a>resource timing buffer full flag</a> to false.
     </dfn> HTTP response header field can be used to communicate a policy
     indicating origin(s) that are allowed to see values of attributes that
     would have been zero due to the cross-origin restrictions. The header's
-    value is represented by the following ABNF [[!RFC5234]]:</p>
+    value is represented by the following ABNF [[!RFC5234]] (using <a href='https://tools.ietf.org/html/rfc7230#section-7'>List Extension</a>, [[!RFC7230]]):</p>
 
     <pre class="abnf">
       Timing-Allow-Origin = 1#( <a href='https://fetch.spec.whatwg.org/#origin-header'>origin-or-null</a> / <a href='https://fetch.spec.whatwg.org/#http-new-header-syntax'>wildcard</a> )


index.html
```diff
@@ -701,7 +701,7 @@
     </dfn> HTTP response header field can be used to communicate a policy
     indicating origin(s) that are allowed to see values of attributes that
     would have been zero due to the cross-origin restrictions. The header's
-    value is represented by the following ABNF [[!RFC5234]]:</p>
+    value is represented by the following ABNF [[!RFC5234]] (using <a href='https://tools.ietf.org/html/rfc7230#section-7'>List Extension</a>, [[!RFC7230]]):</p>
```
igrigorik: I believe we can link directly to 7230 and drop 5234?

plehegar: but rfc 7230 only defines an extension to rfc5234. It doesn't contain the ABNF syntax. So I concluded that linking to both was better than one of them.

igrigorik: Ah, fair enough.

igrigorik:
@plehegar any idea why the travis builds are failing? 

Other then that, LGTM.

plehegar:
we don't have a .travis.yml anymore. I desactivated travis for the time being.

PR #117, opened 2017-09-27
fetchStart is a required field, not a parenthesized zeroed field

### LINKED ISSUE #98: Graphic should indicate fetchStart is a required field, not a parenthesized zeroed field
The [graphic description](http://yoavweiss.github.io/resource-timing/#processing-model) says:

> Attributes in parenthesis indicate that they may not be available if the resource does not pass the timing allow check algorithm.

fetchStart is shown with parenthesis, but it should not be zeroed [according to the timing allow check](http://yoavweiss.github.io/resource-timing/#cross-origin-resources):

> ... these attributes of its PerformanceResourceTiming object must be set to zero: redirectStart, redirectEnd, domainLookupStart, domainLookupEnd, connectStart, connectEnd, requestStart, responseStart, secureConnectionStart, transferSize, encodedBodySize and decodedBodySize.

Likewise the [description of fetchStart](http://yoavweiss.github.io/resource-timing/#dom-performanceresourcetiming-fetchstart) does not mention returning zero:

> On getting, the fetchStart attribute must return as follows:
>
> * The time immediately before the user agent starts to fetch the final resource in the redirection, if there are HTTP redirects or equivalent.
> * The time immediately before the user agent starts to fetch the resource otherwise.

It seems "(fetchStart)" should be "fetchStart" in the graphic.

plehegar:
Fixes #98

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/resource-timing/fetchStart-diagram.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/resource-timing/6cf8efd...54a6b2f.html)


igrigorik:
For future reference, we have the original diagram in: https://docs.google.com/document/d/1I7XGNJ57Qgjkg9pL11s7MK7zGEcwAgdNj1W5f7NKbW8/edit

You can click "edit" and export as SVG. If you don't have edit access, please request it on the doc.

Issue #95, opened 2017-02-10
Clarify resourcetimingbufferfull conditions


JosephPecoraro:
The spec text for firing the resourcetimingbufferfull event is poor:

> To add a PerformanceResourceTiming entry (new entry) in the performance entry buffer, run the following steps:
> 
> 1. If resource timing buffer current size is less than resource timing buffer size limit, run the following substeps:
>       * Add new entry to the performance entry buffer.
>       * increase resource timing buffer current size by 1.
> 2. Otherwise, if the resource timing buffer full flag is false, run the following substeps:
>       * fire a simple event named resourcetimingbufferfull at the Document, with its bubbles attribute initialized to true, and has no default action.
>       * set the resource timing buffer full flag to true.

There are a few issues here.

1. Because of the "Otherwise" at the start of Step 2, my reading of this says that if the limit is 4 and the current size is 3, adding 1 new entry will not dispatch a full event. Existing implementations (WebKit, Chrome,  Firefox) all dispatch a full event as soon as the buffer size reaches the limit.

2. Firing a simple event at the "Document" appears incorrect. It should dispatch at the `performance` object itself. Further there is no Document in Workers.

---

I propose:

1. Drop the word "Otherwise" from Step 2.

2. Change "at the Document" to "at the Performance object"

smaug----:
And why the resourcetimingbufferfull event bubbles if it is supposed to be fired at Performance object.

yoavweiss:
Suggested change makes sense to me. Let's discuss on the call.

Issue #141, opened 2018-01-24
When is resourcetimingbufferfull supposed to fire?


smaug----:
Per spec resourcetimingbufferfull fires synchronously when a PerformanceResourceTiming is added and the buffer is full.
When does that synchronous dispatch actually happen in case of sync XHR for example?
Does the resourcetimingbufferfull fire before or after XHR's readystatechange/load/loadend events?

smaug----:
#95 is also related

smaug----:
In workers the issue applies to importScripts, and there there is the question whether event fires before or after the script has run. I guess it needs to fire before, so that if a script initiates other network connections, resourcetimingbufferfull can have been handled already.

npm1:
Since synchronous XHR is deprecated, I don't think it deserves a new special mention in the spec. That said, for other types the question is still valid. This sounds like a duplicate of https://github.com/w3c/resource-timing/issues/82 and the comments suggest that currently browsers do it before the load event of the resource. We will need to test it and spec it.

smaug----:
synchronous XHR isn't deprecated in workers and resourcetimingbufferfull should fire also in workers.

npm1:
I plan to add some comment saying that entries (and therefore the firing of resourcetimingbufferfull, when applicable) should be added before the load event of a resource. But I'm still confused regarding what needs to be added specifically for sync XHR requests. Notice that things like onreadystatechange should not be used with sync XHR requests https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/onreadystatechange

igrigorik:
> Does the resourcetimingbufferfull fire before or after XHR's readystatechange/load/loadend events?

Per discussion in https://github.com/w3c/resource-timing/issues/82#issuecomment-340076036, we don't make any guarantees on when RT entries are delivered — nor do we want to, I think. As such, resourcetimingbufferfull will fire.. sometime.

@smaug---- close with no action?

smaug----:
The spec really must be more clear.
If entries are delivered async, then it must be defined to happen so, and if buffer then becomes full asynchronously too, it must be defined when the event fires. We can't leave the spec as it is, since it will lead to implementations do different things. Per spec firing resourcetimingbufferfull synchronously during XHR is ok.

@annevk may have some feedback on spec'ing this kind of stuff.

annevk:
[...]

Issue #145, opened 2018-01-31
setResourceTimingBufferSize should not clear "resource timing buffer full flag" when the buffer has shrunk


rniwa:
When `setResourceTimingBufferSize` is called to shrink the resource timing buffer size, it's inappropriate to clear "resource timing buffer full flag" since that would result in every new queuing of a resource timing performance entry to fire resourcetimingbufferfull event even though there was no space made available in the buffer in the last firing of the event.

npm1:
Should setResourceTimingBufferSize fire a resourcetimingbufferfull event if the maxSize parameter is less than or equal to resource timing buffer current size? I see that Chrome currently fires resourcetimingbufferfull in this case (regardless of the flag) but that is not stated anywhere in the spec.

npm1:
I checked that Firefox 58 does not fire the event in this case, and the spec does not prompt this, so I think I should just fix the Chrome behavior.

igrigorik:
Resolved via https://github.com/w3c/resource-timing/pull/146.

Issue #137, opened 2018-01-10
Proposal: API for exposing resource requests


npm1:
We would like to expose in-flight resource requests. The following document is a proposal on how to do this by just using an analogue to PerformanceResourceTiming but for resources requests that have not been completed. We would like to have some feedback on the proposal!

https://docs.google.com/document/d/1NS80J5H796-mbEXHr6uWX3ZJMv-sM-jy9BeZM_yPJ94

npm1:
There's no consensus yet regarding the best way to achieve all of the following points:

- Consistency with other PerformanceEntry objects. These are not modified nor removed from the Performance Timeline.

- Avoid confusing developers regarding which of the entries correspond to in-flight resource requests and which correspond to older entries from requests that have been completed.

- Encourage the use of PerformanceObserver instead of polling, which may involve clearing and causing race conditions if there are multipler pollers.

I've updated the document with the ideas we have so far.

npm1:
Update: latest proposal was https://docs.google.com/presentation/d/1x6QTUdrXtk0faWT1zOTIPdyoKno3WFwhzB3L-mqNYxY/edit#slide=id.p

There were concerns about it:
* We might want to do this from Fetch instead.
* There are use cases where we'd like an observer instead of just polling.

npm1:
We had a discussion with Nic and Charles. Here are the use cases they are considering:

- SPA navigation onload signal: during an SPA nav, developers might want to create an artificial 'onload' signal. They want to know which resources have been started to wait for them, but if none have started after some time then they should abort. This use case requires some nontrivial filtering to avoid counting resources such as 1x1 images, so it might not be possible to satisfy this use case with the initial API. But this API should target eventually being able to handle this use case.

- Allow computing a network quiet window for a Time-To-Interactive metric calculated on users: this is a simple use case where we'd like to keep count of the number of outstanding requests (maybe with some basic filtering) and update as resources begin or end.

- Compute an improved onload signal for the webpage. Since onload is such an unreliable signal, developers could force waiting until certain resources requested have finished being fetched before considering the page to have loaded.

@toddreifsteck does this explanation make sense to you? I think this spells out use cases pretty clearly. The observer pattern is a good fit for all of these use cases assuming the buffered flag and takeRecords() is supported. It does not seem that these use cases require high priority notifications, so even simple integration with PerformanceObserver would work. The polling is a worse fit for these use cases, especially the TTI computation.

[...]

PR #146, opened 2018-01-31
Clarify addition of PerformanceResourceTiming

npm1:
Fixes https://github.com/w3c/resource-timing/issues/95
@tdresser 
@igrigorik 
@marcoscaceres

***
<a href="https://pr-preview.s3.amazonaws.com/npm1/resource-timing/pull/146.html" title="Last updated on Mar 15, 2018, 5:07 AM GMT (dadab1b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/146/38b4edb...npm1:dadab1b.html" title="Last updated on Mar 15, 2018, 5:07 AM GMT (dadab1b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -804,7 +804,7 @@ initially false.</li>
               attribute EventHandler onresourcetimingbufferfull;
 };</pre>
 <p>The <dfn>Performance</dfn> interface is defined in
-[[!PERFORMANCE-TIMELINE-2]].</p>
+[[!HR-TIME-2]].</p>
 <p>The method <dfn>clearResourceTimings</dfn> runs the following
 steps:</p>
 <ol>
@@ -825,7 +825,9 @@ than <a>resource timing buffer current size</a>, no
 <a data-cite=
 '!PERFORMANCE-TIMELINE-2/#dfn-performance-entry-buffer'>performance
 entry buffer</a>.</li>
-<li>Set <a>resource timing buffer full flag</a> to false.</li>
+<li>If the <i>maxSize</i> parameter is greater than <a>resource
+timing buffer current size</a>, set <a>resource timing buffer full
+flag</a> to false.</li>
 </ol>
 <p>The attribute <dfn>onresourcetimingbufferfull</dfn> is the event
 handler for the <dfn>resourcetimingbufferfull</dfn> event described
@@ -852,10 +854,11 @@ substeps:
 <ol style='list-style-type: lower-latin;' data-link-for=
 "Performance">
 <li>Set the <a>resource timing buffer full flag</a> to true.</li>
-<li><a data-cite="!DOM/#concept-event-fire">Fire a simple event</a>
-named <code>resourcetimingbufferfull</code> at the <dfn>Performance</dfn>
-object, with its <code>bubbles</code> attribute initialized to true and
-with no default action.</li>
+<li><a data-cite="!DOM/#concept-event-fire">Fire an event</a>
+named <code>resourcetimingbufferfull</code> at the <a data-cite=
+"!HR-TIME-2/#idl-def-performance">Performance</a> object, with its
+<code>bubbles</code> attribute initialized to true and with no
+default action.</li>
 </ol>
 </li>
 </ol>
@@ -1091,12 +1094,8 @@ the <a>PerformanceResourceTiming</a> object</dfn>.</li>
 "!PERFORMANCE-TIMELINE-2/#dfn-performance-entry-buffer">performance
 entry buffer</a>.</li>
 </ol>
-<p class="issue" data-number="82">This specification does not
-specify whether steps 20 and 21 should run before or after the
-<code>load</code> event of the resource—see <a href=
-"https://github.com/w3c/resource-timing/issues/82">issue 82</a> for
-related discussion.</p>
-<div class="issue" data-number="82"></div>
+<aside class="note">Steps 19 and 20 must be run before the <code>
+load</code> event of the resource.</aside>
 </section>
 <section id="sec-monotonic-clock">
 <h3>Monotonic Clock</h3>
--- a/index.html
+++ b/index.html
@@ -1094,8 +1094,12 @@ the <a>PerformanceResourceTiming</a> object</dfn>.</li>
 "!PERFORMANCE-TIMELINE-2/#dfn-performance-entry-buffer">performance
 entry buffer</a>.</li>
 </ol>
-<aside class="note">Steps 19 and 20 must be run before the <code>
-load</code> event of the resource.</aside>
+<p class="issue" data-number="82">This specification does not
+specify whether steps 20 and 21 should run before or after the
+<code>load</code> event of the resource—see <a href=
+"https://github.com/w3c/resource-timing/issues/82">issue 82</a> for
+related discussion.</p>
+<div class="issue" data-number="82"></div>
 </section>
 <section id="sec-monotonic-clock">
 <h3>Monotonic Clock</h3>
--- a/index.html
+++ b/index.html
@@ -845,15 +845,17 @@ entry buffer</a>.</li>
 <li>Increase <a>resource timing buffer current size</a> by 1.</li>
 </ol>
 </li>
-<li>Otherwise, if the <a>resource timing buffer full flag</a> is
-false, run the following substeps:
+<li>If <a>resource timing buffer current size</a> is greater than
+or equal to <a>resource timing buffer size limit</a> and the
+<a>resource timing buffer full flag</a> is false, run the following
+substeps:
 <ol style='list-style-type: lower-latin;' data-link-for=
 "Performance">
 <li>Set the <a>resource timing buffer full flag</a> to true.</li>
 <li><a data-cite="!DOM/#concept-event-fire">Fire a simple event</a>
-named <code>resourcetimingbufferfull</code> at the Document, with
-its <code>bubbles</code> attribute initialized to true and with no
-default action.</li>
+named <code>resourcetimingbufferfull</code> at the <dfn>Performance</dfn>
+object, with its <code>bubbles</code> attribute initialized to true and
+with no default action.</li>
 </ol>
 </li>
 </ol>


igrigorik:
We have three outstanding issues related to this logic: 

- Clarify resourcetimingbufferfull conditions: https://github.com/w3c/resource-timing/issues/95
- When is resourcetimingbufferfull supposed to fire: https://github.com/w3c/resource-timing/issues/141
- setResourceTimingBufferSize should not clear "resource timing buffer full flag" when the buffer has shrunk: https://github.com/w3c/resource-timing/issues/145

To ensure that we're considering the full picture, could we address these as a single PR?

cc: @JosephPecoraro @smaug---- @rniwa @toddreifsteck

npm1:
Sure sounds good, I'll take a look at the other issues and update this PR.

marcoscaceres:
Drive-by comments 🚙💨

npm1:
Wait where are the comments? :)

By the way I'm not yet sure how to solve issue 141 which is why I haven't updated the PR yet. Maybe I'll add a generic comment saying that the entry should have been added by onload, and this would be accompanied by tests using onload showing that this is currently the case on the major browsers?

index.html
```diff
@@ -845,15 +845,17 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-<li>Otherwise, if the <a>resource timing buffer full flag</a> is
-false, run the following substeps:
+<li>If <a>resource timing buffer current size</a> is greater than
+or equal to <a>resource timing buffer size limit</a> and the
+<a>resource timing buffer full flag</a> is false, run the following
+substeps:
 <ol style='list-style-type: lower-latin;' data-link-for=
 "Performance">
 <li>Set the <a>resource timing buffer full flag</a> to true.</li>
 <li><a data-cite="!DOM/#concept-event-fire">Fire a simple event</a>
```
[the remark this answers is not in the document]
npm1: I removed the word 'simple' because I didn't see it but I also didn't find any recent changes (< 1year) to the fire event algorithm. Is there anything in particular you were thinking about?

index.html
```diff
@@ -845,15 +845,17 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-named <code>resourcetimingbufferfull</code> at the Document, with
-its <code>bubbles</code> attribute initialized to true and with no
-default action.</li>
+named <code>resourcetimingbufferfull</code> at the <dfn>Performance</dfn>
```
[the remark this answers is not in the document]
npm1: Done

npm1:
I've added minimal changes to fix the three issues - please take a look.

index.html
```diff
@@ -1089,12 +1094,8 @@ <h3>Processing Model</h3>
 "!PERFORMANCE-TIMELINE-2/#dfn-performance-entry-buffer">performance
 entry buffer</a>.</li>
 </ol>
-<p class="issue" data-number="82">This specification does not
-specify whether steps 20 and 21 should run before or after the
-<code>load</code> event of the resource—see <a href=
-"https://github.com/w3c/resource-timing/issues/82">issue 82</a> for
-related discussion.</p>
-<div class="issue" data-number="82"></div>
+<aside class="note">Steps 19 and 20 must be run before the <code>
```
igrigorik: This is contrary to the conclusion in https://github.com/w3c/resource-timing/issues/82#issuecomment-340076036. 

Let's resolve https://github.com/w3c/resource-timing/issues/141#issuecomment-364099787 before we land this.

igrigorik:
@JosephPecoraro @rniwa can you please review this to make sure we've addressed #95, #145.

index.html
npm1: Steps 19 and 20 do not involve much work at all and there is no reason for them to not run synchonously. I think what we wanted to delay in Chrome was executing the task in step 6 of https://www.w3.org/TR/performance-timeline-2/#dfn-queue-a-performanceentry
This is actually already spec'd as a low priority task, so Chrome is just trying to actually make it low priority by only running it when idle.

igrigorik: We did not want to encourage use and assumption of entry being available before onload: if the PO entry is delivered asynchronously, we should apply same logic to queuing them into the buffer. Further, if we ever wanted to include processing time of the resource (e.g. https://github.com/w3c/resource-timing/issues/133), locking RT to be queued before onload event would then clash down the road.

npm1: I don't see a problem with having a discrepancy between PO callbacks and entries in buffer - the fact that the callback is async is needed for performance reasons, right? And processing time sounds interesting, but I'm not sure how safe it is to delay pushing the entry by a lot, given that the existing implementations seem to add entries to the buffer ASAP. I'm working on tests to see how browsers behave with several types of resources, but if they all have entries by onload, then I would oppose to specing something that is contrary to that. This is a contentious point so I can remove this 'solution' to issue #141 for now but we should make a decision on this sometime.

index.html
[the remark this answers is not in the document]
marcoscaceres: Apologies, I was recalling in Payment Request we had to formulate the prose for firing the event in a very particular way, but that's because we defined a new event interface. With just using `Event`, I think your current text is all good.

index.html
igrigorik: > I'm working on tests to see how browsers behave with several types of resources

Great. If we have this in time, let's review this on the next call. I'd like to get more eyeballs on this before we determine on how to proceed.

toddreifsteck: Although we want this, we are changing the spec in a way that could break the web. I'd prefer to wait for the tests before we commit this text.

npm1: Here are the tests https://github.com/w3c/web-platform-tests/pull/9459
They pass on Chrome https://chromium-review.googlesource.com/c/chromium/src/+/911868
But the WPT bot seems to be a little broken right now, the failures shown are from script crashing. But at least it runs on Firefox-nightly successfully. So now we just need results from Edge and Safari.

npm1: Now that the WPT bot has been fixed, https://github.com/w3c/web-platform-tests/pull/9459 shows that ResourceTiming is exposed by the onload of a DOM object on all browsers.

yoavweiss: Since it seems this change is web compatible, I'm supportive of it, as it would increase predictability and simplify ResourceTiming based testing, if nothing else.

igrigorik: > Although we want this, we are changing the spec in a way that could break the web. I'd prefer to wait for the tests before we commit this text.

@toddreifsteck to clarify, I don't believe we're breaking anything here.. Existing spec language says the behavior is *not* defined and entries may be added during or after onload. The proposed language is changing that to lock in "before onload". 

> Since it seems this change is web compatible, I'm supportive of it, as it would increase predictability and simplify ResourceTiming based testing, if nothing else.

That's a good argument for adopting this. However, I think this might also make https://github.com/w3c/resource-timing/issues/133 tricky. As of today, it's easy and simple for us to add another timestamp that captures duration of processing / onload time, but if we lock-in the queue step to prior to onload then, well.. We'll need some separate entry.

yoavweiss: > However, I think this might also make #133 tricky

Good point! At the same time, if as a result of https://github.com/w3c/resource-timing/issues/137 we'll end up making the entries mutable (similar to PerformanceNavigationTiming, as discussed on the call) that'd be less of an issue.

igrigorik: True, but that's also rather odd behavior. In this case we're talking about entries available via global timeline, not PerfObserver. You could make the case that we already surface "incomplete" entries via NavTiming, but this just perpetuates the confusion.

yoavweiss: From an analytics provider's perspective, incomplete entries (marked as such, with an event indicating mutation or completion) are better than no entries.

Regardless, this seems like something that all current implementations agree on, so changing this behavior in the future wouldn't be harder than it is right now. (and I suspect it's already not web compatible to change it, but could be proven wrong)

igrigorik: > Regardless, this seems like something that all current implementations agree on, so changing this behavior in the future wouldn't be harder than it is right now.

Fair enough, I think you have me convinced. Let's get a final ack from wider group on our next call, and we can proceed?

yoavweiss: SGTM

Soltik:
It seems that the logic described in section 4.4 does not prevent from missing RT events:

1. If resource timing buffer current size is less than resource timing buffer size limit, run the following substeps:
   a. Add new entry to the performance entry buffer.
   b. increase resource timing buffer current size by 1.
2. Otherwise, if the resource timing buffer full flag is false, run the following substeps:
   a. fire a simple event named resourcetimingbufferfull at the Document, with its bubbles attribute initialized to true, and has no default action.
   b. set the resource timing buffer full flag to true.

Algorithm should include saving of current RT entry:

1. If resource timing buffer current size is less than resource timing buffer size limit-1, run the following substeps:
   a. Add new entry to the performance entry buffer.
   b. increase resource timing buffer current size by 1.
2. Otherwise, if the resource timing buffer full flag is false, run the following substeps:
   a. Add new entry to the performance entry buffer.
   b. fire a simple event named resourcetimingbufferfull at the Document, with its bubbles attribute initialized to true, and has no default action.
   c. set the resource timing buffer full flag to true.

Even simplier way:

1.. If resource timing buffer current size is less than resource timing buffer size limit, run the following substeps:

   a. Add new entry to the performance entry buffer.
   b. increase resource timing buffer current size by 1.
   c. If resource timing buffer current size is equal or greater than resource timing buffer size limit and the resource timing buffer full flag is false, run the following substeps:
   c1. set the resource timing buffer full flag to true.
   c2. fire a simple event named resourcetimingbufferfull at the Document, with its bubbles attribute initialized to true, and has no default action.
   
And another way:
Not to change the algorythm, but to add current RT entry into arguments of event function, so the developer gets at least one lost entry inside callback.

yoavweiss:
> It seems that the logic described in section 4.4 does not prevent from missing RT events:
> 
> If resource timing buffer current size is less than resource timing buffer size limit, run the following substeps:
> a. Add new entry to the performance entry buffer.
> b. increase resource timing buffer current size by 1.
> Otherwise, if the resource timing buffer full flag is false, run the following substeps:
> a. fire a simple event named resourcetimingbufferfull at the Document, with its bubbles attribute initialized to true, and has no default action.
> b. set the resource timing buffer full flag to true.

That's not what this PR actually says. It actually fixes this by removing the "otherwise" and adding an extra condition instead.
So after this PR lands, the processing would be:
<ol>
 <li>If <a>resource timing buffer current size</a> is less than
 <a>resource timing buffer size limit</a>, run the following
 substeps:
 <ol style="list-style-type: lower-latin;">
 <li>Add <i>new entry</i> to the <a data-cite=
 '!PERFORMANCE-TIMELINE-2/#dfn-performance-entry-buffer'>performance
 entry buffer</a>.</li>
 <li>Increase <a>resource timing buffer current size</a> by 1.</li>
 </ol>
 </li>
<li>If <a>resource timing buffer current size</a> is greater than
or equal to <a>resource timing buffer size limit</a> and the
<a>resource timing buffer full flag</a> is false, run the following
substeps:
<ol style='list-style-type: lower-latin;' data-link-for=
 "Performance">
 <li>Set the <a>resource timing buffer full flag</a> to true.</li>
<li><a data-cite="!DOM/#concept-event-fire">Fire an event</a>
named <code>resourcetimingbufferfull</code> at the <a data-cite=
"!HR-TIME-2/#idl-def-performance">Performance</a> object, with its
<code>bubbles</code> attribute initialized to true and with no
 default action.</li>
 </ol>
 </li>

Stepping through this, if the entry we're about to add will reach the buffer's capacity, it will be added (as we'll still pass the "less than" condition), and the counter will be incremented.
At that point, the counter will be equal to the max size, so we'll also fall into the second condition, set the flag to true and fire an event indicating that the buffer is now full.

npm1:
Tim told me that from the discussion in Web Perf WG we would like to land this without the change to require entries to be on the buffer before onload. I didn't quite understand the problem with Safari developer preview, but they said they no longer guarantee entries to be in before onload right @rniwa ? I don't know the reasoning behind this change in Safari but ok. Just want to confirm that I should go ahead and remove the onload part of this PR.

rniwa:
Yes, in WebKit, we don't `resourcetimingbufferfull` event until the next tasks, and once scripts had cleared the buffer, we'd add more entries. This would mean that `load` event for some resources could fire before corresponding resource timing entries are inserted into the buffer.

npm1:
> and once scripts had cleared the buffer, we'd add more entries.
Which entries would you add? If it's the ones added when the resource timing buffer is full, then I think your interpretation of the spec is not correct because the point of having a limit to the buffer size is to limit the amount of memory we spend on storing entries. If you're storing them even when the resource timing buffer is full, then what's the point of having a limit on the size of the buffer? It should also be noted that that implementation makes Safari inconsistent with other browser vendors: Safari will deliver a different amount of resource timing entries than the rest of browsers, which will ignore entries that occur when the buffer is full.

rniwa:
> > and once scripts had cleared the buffer, we'd add more entries.
> Which entries would you add? If it's the ones added when the resource timing buffer is full, then I think your interpretation of the spec is not correct because the point of having a limit to the buffer size is to limit the amount of memory we spend on storing entries.

Perhaps you misunderstood what I'm saying. What we have is a secondary buffer which is used to queue up entries until `resourcetimingbufferfull` event is fired. Once the event is fired and the primary buffer exposed to scripts aren't cleared, then we'd simply drop those entires. We don't indefinitely queue up more entries.

> If you're storing them even when the resource timing buffer is full, then what's the point of having a limit on the size of the buffer?

The point is to be compatible with other browsers which don't drop entries on floor when the primary buffer is full.

> It should also be noted that that implementation makes Safari inconsistent with other browser vendors: Safari will deliver a different amount of resource timing entries than the rest of browsers, which will ignore entries that occur when the buffer is full.

This is absolutely not true. We deliver exactly the same number of resource timing entires as other browsers. We simply delay the dispatching of `resourcetimingbufferfull` event until the next task. While the event is being delayed, we queue up entries in a secondary buffer.

npm1:
Ok, that explanation is a bit clearer. I still think it's a bad idea to have very different implementations (although I guess one benefit of this is encouraging developers to use PerformanceObserver instead?). Safari is delaying the resourcetimingbufferfull event and then trying to simulate the entries as if it had been fired as soon as the resource loaded. To me, that's not intuitive at all and may cause subtle bugs.

For now, I'll remove the onload change and ask for another review.

npm1:
PTAL, reverted onload change.

Issue #110, opened 2017-04-27
InitiatorType of service worker navigation preload


horo-t:
I'm trying to expose the timing information of service worker navigation preload responses.
https://crbug.com/712809

According to the current spec, the initiatorType of the PerformanceResourceTiming of the preload responses must be "other".
But I think it is more convenient if we can use "navigationpreload" or something like it.

igrigorik:
Still trying to wrap my head around implications of "navigation preload", but on first pass.. shouldn't the initiator be "navigation". From what I understand, the navigation proceeds in parallel with the worker startup, so initiator shouldn't change..

yoavweiss:
Related Chromium change (including tests): https://chromium-review.googlesource.com/c/chromium/src/+/941342

yoavweiss:
I tried to add a note, but it seemed out of place, and something that really belonged in Fetch. Looking at Fetch, the "navigation" destination seems to just fall off the definition and a note seemed spurious.

With the tests in place to make sure implementations won't get it wrong in the future, I now feel a note is not really necessary.

Closing, but feel free to reopen if you disagree.

PR #147, opened 2018-02-26
Add a note regarding navigation preload in initiatorType

yoavweiss:
Partially addresses https://github.com/w3c/resource-timing/issues/110 (covers the note, but not the tests)

***
<a href="https://pr-preview.s3.amazonaws.com/yoavweiss/resource-timing/pull/147.html" title="Last updated on Mar 6, 2018, 9:12 AM GMT (59c28ce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/147/38b4edb...yoavweiss:59c28ce.html" title="Last updated on Mar 6, 2018, 9:12 AM GMT (59c28ce)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -450,8 +450,8 @@ method</a> [[!BEACON]];</li>
 <aside data-dfn-for="PerformanceResourceTiming" class="note">
 The "navigation" initiator type includes all types of
 <a data-cite="!FETCH/#navigation-request">navigation requests</a>, including
-<a data-cite=
-"!service-workers-1/#ref-for-service-worker-registration-navigation-preload-enabled-flag%E2%91%A2">
+<a href=
+    "https://w3c.github.io/ServiceWorker/#ref-for-service-worker-registration-navigation-preload-enabled-flag%E2%91%A2">
 navigation preloads</a>.
 </aside>
 <p data-dfn-for="PerformanceResourceTiming">On getting, the
--- a/index.html
+++ b/index.html
@@ -447,6 +447,13 @@ the <a data-cite="!BEACON/#sec-sendBeacon-method">sendBeacon
 method</a> [[!BEACON]];</li>
 <li><dfn>"other"</dfn>, if none of the above conditions match.</li>
 </ul>
+<aside data-dfn-for="PerformanceResourceTiming" class="note">
+The "navigation" initiator type includes all types of
+<a data-cite="!FETCH/#navigation-request">navigation requests</a>, including
+<a data-cite=
+"!service-workers-1/#ref-for-service-worker-registration-navigation-preload-enabled-flag%E2%91%A2">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -447,6 +447,13 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
 method</a> [[!BEACON]];</li>
 <li><dfn>"other"</dfn>, if none of the above conditions match.</li>
 </ul>
+<aside data-dfn-for="PerformanceResourceTiming" class="note">
+The "navigation" initiator type includes all types of
+<a data-cite="!FETCH/#navigation-request">navigation requests</a>, including
+<a data-cite=
+"!service-workers-1/#ref-for-service-worker-registration-navigation-preload-enabled-flag%E2%91%A2">
```
igrigorik: This reference doesn't resolve in SW spec. 

On a related note, should Fetch have this note? Seems like we're correcting for lack of clarify in Fetch spec with respect to navigation preload?

igrigorik: Interesting.. Nope. Also, if I search for "preload" there are no hits in the doc. Hmm?

yoavweiss:
@igrigorik modified to link to be a direct link - PTAL?

index.html
```diff
@@ -447,6 +447,13 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
 method</a> [[!BEACON]];</li>
 <li><dfn>"other"</dfn>, if none of the above conditions match.</li>
 </ul>
+<aside data-dfn-for="PerformanceResourceTiming" class="note">
+The "navigation" initiator type includes all types of
+<a data-cite="!FETCH/#navigation-request">navigation requests</a>, including
+<a href=
+    "https://w3c.github.io/ServiceWorker/#ref-for-service-worker-registration-navigation-preload-enabled-flag%E2%91%A2">
```
igrigorik: "https://w3c.github.io/ServiceWorker/#service-worker-registration-navigation-preload-enabled-flag" seems to work. What's up with the "%E2%91%A2" bit? 

But also coming back to my earlier question: doesn't this note make more sense in Fetch? We're basically clarifying that "navigation requests" should include navigation preload; seems like that should be spelled out in Fetch. 

/cc @annevk

yoavweiss: Looking at the fetch spec, from the definition there ('A navigation request is a request whose destination is "document".') it's obvious that navigation preloads are also included, so I'm not sure a note there makes sense.

Let me know what you think, but I'm willing to drop this note entirely, now that we have tests that make sure future implementers get this right.

igrigorik: If it's obvious, then I don't think we need it? :)

yoavweiss:
Decided on the call that this is not needed and the test is enough.

Issue #89, opened 2016-12-29
Increase limit of 150 for PerformanceEntrys?


nolanlawson:
The spec says:

> The recommended minimum number of PerformanceResourceTiming objects is 150, though this may be changed by the user agent

I tested Chrome 55, Firefox 50, Edge 14, and IE 11, and found that Firefox was the only one to limit all entries by default to 150. For the others, I added up to 100,000 entries and saw no limit. ([Here's a demo](http://bl.ocks.org/nolanlawson/e686c8e42c3f2fa7eecd031263053004)).

150 seems like an extremely low limit. Consider my project [marky](https://github.com/nolanlawson/marky), which attempts to provide a sugar API on top of `performance.measure()`, but because it adds two marks and a measure (I do this for maximum debuggability in Chrome Dev Tools, Edge F12, and Windows Performance Analyzer), it's effectively limited to 50 entries in Firefox.

Of course you can increase the limit with `setResourceTimingBufferSize()`, but this still seems like it could cause compatibility issues for Firefox, since many developers may skip this step. Maybe it would be prudent to increase the limit in the spec itself, to something more forgiving like 10,000?

igrigorik:
> Maybe it would be prudent to increase the limit in the spec itself, to something more forgiving like 10,000?

We could, but I also think we want to encourage use of PerformanceObserver's instead of perpetuating the buffer model.. a lower cap on the buffer is a good incentive to switch. That said, we won't be able to eliminate the buffer entirely either -- see https://github.com/w3c/performance-timeline/issues/56#issuecomment-268922807. My proposal would be to evaluate the limit with respect to "is it large enough to buffer records emitted for before onload event has finished executed", and after that we want to nudge folks to use PerfObserver.

Looking at the latest HTTP Archive crawl...

![image](https://cloud.githubusercontent.com/assets/10652/21550950/b33764ac-cdcb-11e6-91e6-17127c04a4f7.png)

^ does suggest that we should consider raising it.

nolanlawson:
> We could, but I also think we want to encourage use of PerformanceObserver's instead of perpetuating the buffer model…

[...]

PR #155, opened 2018-04-22
Update recommended min UA Resource Timing buffer size to 250

nicjansma:
I'd like to propose a modest change to the "recommended" initial UA RT buffer size of 150.

I've been doing some [research](https://nicj.net/resourcetiming-visibility-third-party-scripts-ads-and-page-weight/) into various aspects of ResourceTiming "visibility", and have found that over 14% of sites generate more than 150 resources in their main frame.

Here's a distribution of the Alexa Top 1000 sites and how many RT entries are from their main frame (from a [crawl](https://github.com/nicjansma/resourcetiming-visibility) with the buffer size increased in the head):

image

In this crawl, ~18% of sites would have hit the 150 limit (and only 15% of _those_ sites set the buffer size higher).

The 150 number is a UA-recommended limit, though I believe all current UAs choose 150.  It is not a spec requirement.

There's some previous discussion on whether 150 is appropriate, or if we should suggest UAs pick a higher default here: https://github.com/w3c/resource-timing/issues/89

The reason I think it should be increased is that:

* IIRC, we [initially chose](https://lists.w3.org/Archives/Public/public-web-perf/2012Apr/0034.html) 150 "back in the day" because many sites were nearing 100 resources or more, and we thought 150 would give enough buffer room to track all of those resources for the majority of sites
* Today, the [HTTPArchive](https://httparchive.org/reports/state-of-the-web#reqTotal) suggests a median of 80 resources and 90th percentile of 205 resources, but it's important to note that this is per-page, not per-frame.  I think it's safe to say the request count per page has increased since 2012, so it's a good time to re-evaluate the recommended minimum.
* While a site can update their buffer size beyond 150 by simply calling `setResourceTimingBufferSize`, third-party analytics scripts that may be loaded lower priority or after onload might want to increase this limit to ensure they capture all of the page's resources, but by the time they load, the buffer may already be full

It would still be up to UAs to decide if they want to chose 250, stay at 150, or a go for a lower initial limit (maybe even depending on the environment).

Changing to 250 will leave only 2.5% of the Alexa Top 1000 sites with a full buffer (more than 250 resources in the main frame).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/pull/155.html" title="Last updated on May 30, 2018, 1:09 AM GMT (3a0db55)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/155/ea8a875...3a0db55.html" title="Last updated on May 30, 2018, 1:09 AM GMT (3a0db55)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -784,7 +784,7 @@ Timeline</a> [[!PERFORMANCE-TIMELINE-2]]. This section extends the
 interface to allow controls over the number of
 <a>PerformanceResourceTiming</a> objects stored.</p>
 <p>The recommended minimum number of
-<a>PerformanceResourceTiming</a> objects is 150, though this may be
+<a>PerformanceResourceTiming</a> objects is 250, though this may be
 changed by the user agent. <a data-link-for=
 "Performance">setResourceTimingBufferSize</a> can be called to
 request a change to this limit.</p>
@@ -792,7 +792,7 @@ request a change to this limit.</p>
 environment</a> has:</p>
 <ul>
 <li>a <dfn>resource timing buffer size limit</dfn> which should
-initially be 150 or greater.</li>
+initially be 250 or greater.</li>
 <li>a <dfn>resource timing buffer current size</dfn> which is
 initially 0.</li>
 <li>a <dfn>resource timing buffer full flag</dfn> which is


index.html
```diff
@@ -784,15 +784,15 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
 interface to allow controls over the number of
 <a>PerformanceResourceTiming</a> objects stored.</p>
 <p>The recommended minimum number of
-<a>PerformanceResourceTiming</a> objects is 150, though this may be
+<a>PerformanceResourceTiming</a> objects is 250, though this may be
```
yoavweiss: Slightly tangential nit: I think we need to define a stronger language for the minimum, as I suspect that UAs which will decide to go with lower buffer sizes will be ignored by developers, which will result in dropped entries and potential compat issues. That doesn't have to happen in this PR though.

yoavweiss:
I support this change, but I think it needs some commitment from all UAs, as partial adoption may lead to dropped entries (when developers start assuming that the new value is the minimum everywhere)

cc @igrigorik @tdresser @toddreifsteck @rniwa @digitarald

tdresser:
This seems sane to me - just a quick question about when we buffer events.
In PR [95 of the Performance Timline spec](https://github.com/w3c/performance-timeline/pull/95/files#r143523605), there's discussion of only buffering up to onload.

Are we planning on having that apply to resource timing? If so, do we know what the data looks like if we only look at resources fetched before onload?

igrigorik:
@tdresser I don't think we'll be able to stop queuing RT entries post-onload once we land the PO buffering logic, as that would break nearly every existing client. RT will have to be a special case where we do both: we will provide a way for PO to retrieve buffered entries queued prior to onload (e.g. via passing buffered flag to PO registration), and we will continue the logging post onload into the existing buffer. For new specs, we don't have to do this, but given wide adoption of RT I think we need to provide both to make the migration path simple, without breaking old clients.

FWIW, I think the "what is the distribution of same-frame resources prior to onload" question is worth investigating a bit further. I believe we should have all the right bits in HTTP Archive to extract this: we have detailed data on initiator / frame, and we can filter out any requests initiated after onload fires. @nicjansma @rviscomi perhaps you can do some digging?

toddreifsteck:
LGTM

yoavweiss:
@nicjansma can you open implementation bugs to modify that value and link them here?
Since this is not a requirement, we can't really test it, but at the very least we need issue to track it, otherwise I suspect implementations may not follow.

rviscomi:
> FWIW, I think the "what is the distribution of same-frame resources prior to onload" question is worth investigating a bit further. I believe we should have all the right bits in HTTP Archive to extract this: we have detailed data on initiator / frame, and we can filter out any requests initiated after onload fires. @nicjansma @rviscomi perhaps you can do some digging?

Happy to help, but not sure of the exact problem statement. It may be worth creating a new thread on discuss.httparchive.org and documenting the question/answer so the community can help as well.

past:
Firefox 62 will contain this change per https://bugzilla.mozilla.org/show_bug.cgi?id=1462880

nicjansma:
Filed https://bugs.chromium.org/p/chromium/issues/detail?id=847689 for Chrome

yoavweiss:
Chromium CL is at https://chromium-review.googlesource.com/c/chromium/src/+/1078107

yoavweiss:
> FWIW, I think the "what is the distribution of same-frame resources prior to onload" question is worth investigating a bit further. I believe we should have all the right bits in HTTP Archive to extract this: we have detailed data on initiator / frame, and we can filter out any requests initiated after onload fires. @nicjansma @rviscomi perhaps you can do some digging?

@igrigorik - just noticed that the "what is the distribution of same-frame resources prior to onload" question went unanswered.

As far as cross-origin iframes go, their data was not included in @nicjansma's research, as RUM has no visibility into their perf timeline.

So the HTTPArchive analysis will enable us to see if there's a huge difference between the same-origin's main frame and same-origin iframes as far as resource numbers go, but I doubt that would significantly reduce the numbers in most cases. I agree that it seems worthwhile to dig further there, but not sure we should block on it. WDYT?

igrigorik:
> So the HTTPArchive analysis will enable us to see if there's a huge difference between the same-origin's main frame and same-origin iframes as far as resource numbers go, but I doubt that would significantly reduce the numbers in most cases. I agree that it seems worthwhile to dig further there, but not sure we should block on it. WDYT?

You're probably right, but I'd still love to sanity check this with data. :)

> Happy to help, but not sure of the exact problem statement. It may be worth creating a new thread on discuss.httparchive.org and documenting the question/answer so the community can help as well.

The specific question we're after is: what is the distribution of the count of resources fetched by pages, where the fetch is initiated prior to end of the onload event of the page — e.g. excluding resources deferred to post onload, lazyloaded content, etc.

rviscomi:
Ran the HTTP Archive analysis on the distribution of requests before onload. See the full writeup here: https://discuss.httparchive.org/t/requests-per-page-before-onload/1376

In summary, the median is 74 requests, with a bimodal distribution peaking around 1 and 50 requests.

image

igrigorik:
@rviscomi awesome work, thank you! 

> IIUC did you also want to split this by initiator?

I don't think it would directly affect what we're after here, but might be an interesting datapoint.

One thing, what percentile is 250 req's? What are the 95, 97, 99th, etc?

rviscomi:
250 is the 95.86th %ile

The 95th is 234
The 97th is 280
The 99th is 415

For more detailed percentile info see the raw data in [this sheet](https://docs.google.com/spreadsheets/d/16x-4uO90MZYxa6WtuOjZag-_aVsrA-dln0jjmVeYcdM/edit?usp=sharing).

yoavweiss:
@rviscomi awesome data, thanks! :)
Generally, it seems like the RUM data and HA data more or less agree and 250 resources sits between the 95th and the 98th %tile.

Since it all adds up, I haven't heard objections and this change has already landed in 2 implementations - I'm merging it.

PR #159, opened 2018-05-18
Add nextHopProtocol HTTP/0.9 and 1.0. Link to QUIC

### LINKED ISSUE #122: NextHopProtocol Values
I'm investigating the nextHopProtocol (https://w3c.github.io/resource-timing/#dom-performanceresourcetiming-nexthopprotocol), and I need to know what possible values it can have. According to ALPN Protocol ID [RFC7301] (https://tools.ietf.org/html/rfc7301#section-6), it appears that the set of possible values is "http/1.1", "spdy/1", "spdy/2", and "spdy/3". To test this, I went on several websites and examined the nextHopProtocol value by entering performance.getEntriesByType('navigation')[0].nextHopProtocol into the JS console. I found some instances of "http/1.1" (e.g. on https://go.twitch.tv/, https://www.apple.com/), but I also found "hq" (e.g. on https://www.google.com/, the new tab page) and "h2" (e.g. on https://twitter.com, https://www.reddit.com/).

From the spec, it's not clear what "hq" or "h2" mean. It would also be beneficial to have a master list of all possible values that the nextHopProtocol can be.

And for the record, I did this investigation on a Mac using Chrome Version 63.0.3230.0 (Official Build) dev (64-bit).

yoavweiss:
Closes #122

***
<a href="https://pr-preview.s3.amazonaws.com/yoavweiss/resource-timing/pull/159.html" title="Last updated on Jun 15, 2018, 9:59 AM GMT (de148c0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/159/0b6e799...yoavweiss:de148c0.html" title="Last updated on Jun 15, 2018, 9:59 AM GMT (de148c0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -469,10 +469,13 @@ protocols, which are not defined in the ALPN Protocol ID registry,
 user agents MAY return "http/0.9" and "http/1.0" respectively. In
 case the user agent is using an experimental protocol which ALPN
 protocol ID has not yet been defined in the above registry, the user
-agent MAY use another descriptive string.</p>
+agent MAY use the ALPN negotiated value or use another descriptive
+string.</p>
 
-<p class=note>The "hq" ALPN ID is defined for the QUIC protocol in the <a
-   href="https://tools.ietf.org/html/draft-ietf-quic-http-11#section-9.1">HTTP over QUIC Internet Draft</a>.
+<p class=note>The "hq" ALPN ID is defined for the final version of
+the HTTP over QUIC protocol in the
+<a href="https://tools.ietf.org/html/draft-ietf-quic-http-11#section-9.1">
+HTTP over QUIC Internet Draft</a>.
 </p>
 <p>Note that the <a data-link-for=
 "PerformanceResourceTiming">nextHopProtocol</a> attribute is
--- a/index.html
+++ b/index.html
@@ -455,7 +455,10 @@ used to fetch the resource, as identified by the ALPN Protocol ID
 or local resources, return an empty string. The registry for
 protocol identifiers lists the possible values for the <a href=
 "https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids">
-ALPN Protocol IDs</a>. When a proxy is configured, if a tunnel
+ALPN Protocol IDs</a>. For the HTTP/0.9 and HTTP/1.0 network protocols, which are not
+defined in the ALPN Protocol ID registry, user agents MAY return "http/0.9" and
+"http/1.0" respectively.
+When a proxy is configured, if a tunnel
 connection is established then this attribute MUST return the ALPN
 Protocol ID of the tunneled protocol, otherwise it MUST return the
 ALPN Protocol ID of the first hop to the proxy. In order to have
@@ -464,6 +467,9 @@ additional constraints apply: octets in the ALPN protocol MUST NOT
 be percent-encoded if they are valid token characters except "%",
 and when using percent-encoding, uppercase hex digits MUST be
 used.</p>
+<p class=note>The "hq" ALPN ID is defined for the QUIC protocol in the <a
+   href="https://tools.ietf.org/html/draft-ietf-quic-http-11#section-9.1">HTTP over QUIC Internet Draft</a>.
+</p>
 <p>Note that the <a data-link-for=
 "PerformanceResourceTiming">nextHopProtocol</a> attribute is
 intended to identify the network protocol in use for the fetch
--- a/index.html
+++ b/index.html
@@ -452,21 +452,25 @@ attribute <dfn>nextHopProtocol</dfn> returns the network protocol
 used to fetch the resource, as identified by the ALPN Protocol ID
 [[!RFC7301]]; resources retrieved from <a data-cite=
 "!HTML/#relevant-application-cache">relevant application caches</a>
-or local resources, return an empty string. The registry for
-protocol identifiers lists the possible values for the <a href=
+or local resources, return an empty string. When a proxy is
+configured, if a tunnel connection is established then this attribute
+MUST return the ALPN Protocol ID of the tunneled protocol, otherwise
+it MUST return the ALPN Protocol ID of the first hop to the proxy.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -464,6 +467,9 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
 be percent-encoded if they are valid token characters except "%",
 and when using percent-encoding, uppercase hex digits MUST be
 used.</p>
+<p class=note>The "hq" ALPN ID is defined for the QUIC protocol in the <a
+   href="https://tools.ietf.org/html/draft-ietf-quic-http-11#section-9.1">HTTP over QUIC Internet Draft</a>.
```
igrigorik: We recently changed this:
- https://github.com/w3c/navigation-timing/issues/71
- background: https://lists.w3.org/Archives/Public/ietf-http-wg/2018JanMar/0204.html 

We should probably capture the resolution from that discussion in this note? 

Otherwise, LGTM!

yoavweiss: Theoretically, that has not changed, it's just that Chrome is using a different string for G-QUIC, instead of publishing it as the IETF variant, which "owns" the ALPN ID.

I wonder if we need to also add some liberty on that front for experimental protocols. Maybe something like: "In case the network protocol used is an experimental one and not defined in the above registry, user agents MAY also use other descriptive strings."

igrigorik: Yeah, that makes sense. In fact, perhaps invert it and lead with note about experimental protocols, citing QUIC as one example.

yoavweiss: I think the bit about experimental protocols needs to be normative. Maybe I can add it around the HTTP/0.9 and 1.0 bits. (I'm also thinking it'd make for a long paragraph, so maybe I should split it up into a list somehow)

yoavweiss: (belatedly) added the above. PTAL

index.html
```diff
@@ -469,8 +469,8 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
 user agents MAY return "http/0.9" and "http/1.0" respectively. In
 case the user agent is using an experimental protocol which ALPN
 protocol ID has not yet been defined in the above registry, the user
-agent MAY use the ALPN negotiated value or use another descriptive
-string.</p>
+agent SHOULD use the ALPN negotiated value or MAY use another descriptive
```
igrigorik: Do we really both a SHOULD and a MAY here? Can we collapse them?

> user agent SHOULD use the ALPN negotiated value or another descriptive string if the ALPN negotiated value is not available.

index.html
```diff
@@ -452,18 +452,31 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
...
+or local resources, return an empty string. When a proxy is
+configured, if a tunnel connection is established then this attribute
+MUST return the ALPN Protocol ID of the tunneled protocol, otherwise
+it MUST return the ALPN Protocol ID of the first hop to the proxy.
+In order to have precisely one way to represent any ALPN protocol ID,
+the following additional constraints apply: octets in the ALPN
+protocol MUST NOT be percent-encoded if they are valid token
+characters except "%", and when using percent-encoding, uppercase hex
+digits MUST be used.</p>
+<p>The registry for protocol identifiers lists the possible values
```
MikeBishop: The whole point of the issue is that the registry isn't the set of "possible" values, it's a set of values whose meaning is generally well-defined.  I'd say that the registry lists some common values, but implementations need to be prepared to encounter identifiers for experimental or unregistered protocols.

index.html
```diff
@@ -452,18 +452,31 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
...
-Protocol ID of the tunneled protocol, otherwise it MUST return the
-ALPN Protocol ID of the first hop to the proxy. In order to have
-precisely one way to represent any ALPN protocol ID, the following
-additional constraints apply: octets in the ALPN protocol MUST NOT
-be percent-encoded if they are valid token characters except "%",
-and when using percent-encoding, uppercase hex digits MUST be
-used.</p>
+ALPN Protocol IDs</a>. For the HTTP/0.9 and HTTP/1.0 network
+protocols, which are not defined in the ALPN Protocol ID registry,
+user agents MAY return "http/0.9" and "http/1.0" respectively. In
```
MikeBishop: Just submitted the registration request for these, so hopefully this sentence will become obsolete next week.

index.html
```diff
@@ -452,18 +452,31 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
...
+case the user agent is using an experimental protocol which ALPN
+protocol ID has not yet been defined in the above registry, the user
+agent SHOULD use the ALPN negotiated value or MAY use another descriptive
+string if one is not available.</p>
```
MikeBishop: As suggested on the issue:  MUST use the negotiated ALPN token if any, MAY use another descriptive string if ALPN was not used.

yoavweiss: Changed in that spirit. PTAL

index.html
yoavweiss: Removed for now in hope it will

yoavweiss:
OK, I believe I addressed all the comments.
@igrigorik @MikeBishop @lpardue @ryanatgoogle PTAL?

LPardue:
Thanks. I wonder, in light of removing explicit HTTP 0.9 and 1.0 and adding  negotiation text, could you be less explicit about the "hq" identifier? Perhaps just point to the IETF HTTP/QUIC document, which already fully defines how to handle ALPN identifiers.

yoavweiss:
> Thanks. I wonder, in light of removing explicit HTTP 0.9 and 1.0 and adding negotiation text, could you be less explicit about the "hq" identifier? Perhaps just point to the IETF HTTP/QUIC document, which already fully defines how to handle ALPN identifiers.

The reason this is added (as a non-normative note) is that it already caused some confusion when folks tried to find "hq"'s definition. I don't see harm in us being explicit about it (in a note).

LPardue:
I can live with that 😀

Issue #96, opened 2017-02-10
Clarify behavior of setResourceTimingBufferSize


JosephPecoraro:
[Spec](https://w3c.github.io/resource-timing/#dom-performance-setresourcetimingbuffersize) for setResourceTimingBufferSize is very simple:

> The setResourceTimingBufferSize method runs the following steps:
>
> 1. Set resource timing buffer size limit to the maxSize parameter. If the maxSize parameter is less than resource timing buffer current size, no PerformanceResourceTiming objects are to be removed from the performance entry buffer.
> 2. Set resource timing buffer full flag to false.

However, most implementations do not appear to matches this.

Do we want to change the spec or should I just file bugs?

---

Test case:

```js
function loadRandomResource() {
    return fetch(window.location.href + "?" + Math.random());
}

// Get the browser into a consistent state.
performance.clearResourceTimings();
performance.setResourceTimingBufferSize(100);

// Gather up 3 Resource Entries to kick off the rest of test behavior.
let observer = new PerformanceObserver(function(list) {
    let allResourceEntries = performance.getEntriesByType("resource");
    if (allResourceEntries.length !== 3)
        return;
    observer.disconnect();
    test();
});
observer.observe({entryTypes: ["resource"]});
for (let i = 0; i < 3; ++i)
    loadRandomResource();

// Test behavior of setting buffer limit to the current size (3).
function test() {
    performance.onresourcetimingbufferfull = function() {
        console.log("Event Fired", performance.getEntriesByType("resource").length);
    };
    console.log("before setLimit(3)");
    performance.setResourceTimingBufferSize(3);
    console.log("after setLimit(3)");
    loadRandomResource().then(function() {
        console.log("after loading 4th resource");
    });
}
```

* Expected:

    before setLimit(3)
    after setLimit(3)
    Event Fired - 3
    after loading 4th resource

* Chrome dispatches the buffer full event inside setResourceTimingBufferSize if the new limit is less than or equal to the current size. [(source)](https://chromium.googlesource.com/chromium/blink/+/master/Source/core/timing/PerformanceBase.cpp#167)

    before setLimit(3)
    Event Fired - 3
    after setLimit(3)

* Firefox doesn't maintain separate limit and "buffer full flag". So when the 4th resource loads it is already over the limit so it doesn't dispatch an event. [(source)](https://hg.mozilla.org/mozilla-central/file/tip/dom/performance/Performance.cpp)

    before setLimit(3)
    after setLimit(3)
[...]

PR #163, opened 2018-08-30
Fire `resourcetimingbufferfull` asynchronously

### LINKED ISSUE #141: When is resourcetimingbufferfull supposed to fire?

yoavweiss:
Fixes #141 and #96 

This aligns the spec with WebKit's implementation to make sure that `resourcetimingbufferfull` can fire asynchronously without losing any entries in the process.

Tests are still missing.

@igrigorik @npm1 - can you take a look?

***
<a href="https://pr-preview.s3.amazonaws.com/yoavweiss/resource-timing/pull/163.html" title="Last updated on Sep 1, 2018, 8:20 PM GMT (059e4af)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/163/709c54a...yoavweiss:059e4af.html" title="Last updated on Sep 1, 2018, 8:20 PM GMT (059e4af)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -869,7 +869,7 @@ entry buffer</a>.</li>
 substeps:
 <ol style="list-style-type: lower-latin;">
 <li>Set <a>resource timing buffer full flag</a> to true.</li>
-<li>Queue a task to run <a>fire a buffer full event</a></li>
+<li>Queue a task to run <a>fire a buffer full event</a>.</li>
 </ol>
 <li>If <a>resource timing secondary buffer current size</a> is less than
 <a>resource timing buffer size limit</a>, add <i>new entry</i> to the
@@ -883,20 +883,26 @@ current size</a> by 1.</li>
 named <code>resourcetimingbufferfull</code> at the <a data-cite=
 "!HR-TIME-2/#idl-def-performance">Performance</a> object, with its
 <code>bubbles</code> attribute initialized to true.</li>
-<li>If <a>resource timing buffer size limit</a> minus <a>resource timing buffer
-current size</a> is not larger than <a>resource timing secondary buffer current
-size</a>, abort these steps.</li>
-<p class=note>This means that if the buffer does not have enough room for all
-the entries in the secondary buffer, none of them will be copied.</p>
-<li>For each <i>entry</i> in <a>secondary entry buffer</a>, add <i>entry</i> to
-<a data-cite='!PERFORMANCE-TIMELINE-2/#dfn-performance-entry-buffer'>performance
+<li>While <a>resource timing buffer current size</a> is smaller than
+<a>resource timing buffer size limit</a>, run the following substeps:
+<ol>
+<li>Let <i>entry</i> be the oldest <a>PerformanceResourceTiming</a> in
+<a>secondary entry buffer</a>.</li>
+<li>Add <i>entry</i> to <a
+data-cite='!PERFORMANCE-TIMELINE-2/#dfn-performance-entry-buffer'>performance
 entry buffer</a>.</li>
+<li>Increment <a>resource timing buffer current size</a> by 1.</li>
+</ol>
+<p class=note>This means that if the buffer does not have enough room for all
+the entries in the secondary buffer, some of them will not be copied.
+Developers should make sure that <code>resourcetimingbufferfull</code> event
+handlers call <code>clearResourceTimings</code> or extend the buffer
+sufficiently (by calling <code>setResourceTimingBufferSize</code>).</p>
 <li>Remove all objects from <a>secondary entry buffer</a>.</li>
-<li>Set <a>resource timing buffer current size</a> to <a>resource timing
-secondary buffer current size</a>.</li>
 <li>Set <a>resource timing secondary buffer current size</a> to 0.</li>
 </ol>
 </section>
+
 <section id="sec-cross-origin-resources">
 <h3>Cross-origin Resources</h3>
 <p data-link-for="PerformanceResourceTiming">Cross-origin resources
--- a/index.html
+++ b/index.html
@@ -797,15 +797,20 @@ interface to allow controls over the number of
 changed by the user agent. <a data-link-for=
 "Performance">setResourceTimingBufferSize</a> can be called to
 request a change to this limit.</p>
+
 <p>Each <a data-cite="!WEBIDL#es-environment">ECMAScript global
 environment</a> has:</p>
 <ul>
-<li>a <dfn>resource timing buffer size limit</dfn> which should
+<li>A <dfn>resource timing buffer size limit</dfn> which should
 initially be 250 or greater.</li>
-<li>a <dfn>resource timing buffer current size</dfn> which is
+<li>A <dfn>resource timing buffer current size</dfn> which is
 initially 0.</li>
-<li>a <dfn>resource timing buffer full flag</dfn> which is
+<li>A <dfn>resource timing buffer full flag</dfn> which is
 initially false.</li>
+<li>A <dfn>resource timing secondary buffer current size</dfn> which is
+initially 0.</li>
+<li>A <dfn>secondary entry buffer</dfn> to store <a>PerformanceEntry</a>
+objects that is initially empty.</li>
 </ul>
 <pre class="idl">partial interface Performance {
   void clearResourceTimings ();
@@ -817,13 +822,14 @@ initially false.</li>
 <p>The method <dfn>clearResourceTimings</dfn> runs the following
 steps:</p>
 <ol>
-<li>remove all <a>PerformanceResourceTiming</a> objects in the
+<li>Remove all <a>PerformanceResourceTiming</a> objects in the
 <a data-cite=
 '!PERFORMANCE-TIMELINE-2/#dfn-performance-entry-buffer'>performance
 entry buffer</a>.</li>
-<li>set <a>resource timing buffer current size</a> to 0.</li>
-<li>set <a>resource timing buffer full flag</a> to false.</li>
+<li>Set <a>resource timing buffer current size</a> to 0.</li>
+<li>Set <a>resource timing buffer full flag</a> to false.</li>
 </ol>
+
 <p>The <dfn>setResourceTimingBufferSize</dfn> method runs the
 following steps:</p>
 <ol>
@@ -838,14 +844,16 @@ entry buffer</a>.</li>
 timing buffer current size</a>, set <a>resource timing buffer full
 flag</a> to false.</li>
 </ol>
+
 <p>The attribute <dfn>onresourcetimingbufferfull</dfn> is the event
 handler for the <dfn>resourcetimingbufferfull</dfn> event described
 below.</p>
-<p>To <dfn>add a PerformanceResourceTiming entry</dfn> (<i>new
-entry</i>) in the <a data-cite=
+
+<p>To <dfn>add a PerformanceResourceTiming entry</dfn> into the <a data-cite=
 '!PERFORMANCE-TIMELINE-2/#dfn-performance-entry-buffer'>performance
 entry buffer</a>, run the following steps:</p>
 <ol>
+<li>Let <i>new entry</i> be the input <a>PerformanceEntry</a> to be added.
 <li>If <a>resource timing buffer current size</a> is less than
 <a>resource timing buffer size limit</a>, run the following
 substeps:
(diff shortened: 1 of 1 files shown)


yoavweiss:
@josephpecoraro @rniwa - can you take a look as well?

index.html
```diff
@@ -854,21 +862,39 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-or equal to <a>resource timing buffer size limit</a> and the
-<a>resource timing buffer full flag</a> is false, run the following
+<li>If <a>resource timing buffer full flag</a> is false, run the following
 substeps:
-<ol style='list-style-type: lower-latin;' data-link-for=
-"Performance">
-<li>Set the <a>resource timing buffer full flag</a> to true.</li>
+<ol style="list-style-type: lower-latin;">
+<li>Set <a>resource timing buffer full flag</a> to true.</li>
+<li>Queue a task to run <a>fire a buffer full event</a></li>
```
npm1: nit: period missing

index.html
```diff
@@ -854,21 +862,39 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
+<p>To <dfn>fire a buffer full event</dfn>, run the following steps:</p>
+<ol>
 <li><a data-cite="!DOM/#concept-event-fire">Fire an event</a>
 named <code>resourcetimingbufferfull</code> at the <a data-cite=
 "!HR-TIME-2/#idl-def-performance">Performance</a> object, with its
 <code>bubbles</code> attribute initialized to true.</li>
-</ol>
-</li>
+<li>If <a>resource timing buffer size limit</a> minus <a>resource timing buffer
+current size</a> is not larger than <a>resource timing secondary buffer current
```
npm1: larger than or equal to?

index.html
```diff
@@ -854,21 +862,39 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
+size</a>, abort these steps.</li>
```
npm1: I think it might make sense to clear the secondary buffer even in the 'abort' case, so that we can keep listening to future entries even if the primary buffer was not cleared during this call. This is in line with the previous behavior, which would ignore entries when the buffer is full and the buffer full flag is true. Or even better but more complex, make the secondary buffer a queue, and if the queue becomes full and we want to add an entry then we pop the first in line (oldest entry). This way we maintain a list of 'recent' entries ready for when the primary buffer becomes available. WDYT? Also, I think we should add some entries from the secondary buffer even if not all of them can be added.

index.html

index.html
yoavweiss: good catch

index.html
yoavweiss: I think it makes sense to clear the secondary buffer either way.
I don't know if adding more complexity here (e.g. making it a queue) would provide a lot of benefit.

I guess we can add some entries from secondary to primary even if not all of them can fit, as long as we don't trigger RTbufferfull from within "fire a buffer full event"

npm1:
Thanks, LGTM! I should note a non-obvious change on this PR (not sure it's intentional?):

Before, setResourceTimingBufferSize would never fire resourcetimingbufferfull, and resourcetimingbufferfull would be fired immediately once the buffer becomes full when adding entries (if size is 3, fired after adding the 3rd entry). But this is buggy: if I setResourceTimingBufferSize to reduce the limit to something less than the current size then no event is fired, but then the next time we try to add an entry, there's no space so we fire the event but we are forced to drop the entry. I think it is for this reason that Chrome fires the event on setResourceTimingBufferSize when needed (avoids forcefully dropping the next entry).

With this PR, resourcetimingbufferfull is only fired (or more precisely, now a task is queued to fire it) when we try to add an entry and fail to do so (if size is 3, fired when failing to add the 4th entry). In particular it is still not fired on setResourceTimingBufferSize. But now doing this is fine because the secondary buffer should generally prevent us from dropping the entries. So this fixes the bug :)

yoavweiss:
> Thanks, LGTM! I should note a non-obvious change on this PR (not sure it's intentional?):
> 
> Before, setResourceTimingBufferSize would never fire resourcetimingbufferfull, and resourcetimingbufferfull would be fired immediately once the buffer becomes full when adding entries (if size is 3, fired after adding the 3rd entry). But this is buggy: if I setResourceTimingBufferSize to reduce the limit to something less than the current size then no event is fired, but then the next time we try to add an entry, there's no space so we fire the event but we are forced to drop the entry. I think it is for this reason that Chrome fires the event on setResourceTimingBufferSize when needed (avoids forcefully dropping the next entry).
> 
> With this PR, resourcetimingbufferfull is only fired (or more precisely, now a task is queued to fire it) when we try to add an entry and fail to do so (if size is 3, fired when failing to add the 4th entry). In particular it is still not fired on setResourceTimingBufferSize. But now doing this is fine because the secondary buffer should generally prevent us from dropping the entries. So this fixes the bug :)

This change is indeed intentional. Thanks for reviewing! :)

yoavweiss:
Waiting on tests before merging

npm1:
I think the buffer mechanic is independent on the fact that this is resource timing so it might be worth generalizing. The only possible difference is whether we are forced to trigger the event async or not, but I think it would be better to be consistent and do async always even if not needed for a particular entryType. That said, I wonder if we really do need the clearing buffer mechanism for entry types that are only buffered until load. It seems to me that we want to stop buffering after load, so not sure how important this is in that case.

index.html
```diff
@@ -854,23 +862,49 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
 substeps:
-<ol style='list-style-type: lower-latin;' data-link-for=
-"Performance">
-<li>Set the <a>resource timing buffer full flag</a> to true.</li>
+<ol style="list-style-type: lower-latin;">
+<li>Set <a>resource timing buffer full flag</a> to true.</li>
+<li>Queue a task to run <a>fire a buffer full event</a>.</li>
+</ol>
+<li>If <a>resource timing secondary buffer current size</a> is less than
+<a>resource timing buffer size limit</a>, add <i>new entry</i> to the
```
rniwa: We don't check condition in our implementation since the script wouldn't have a chance to clear the buffer until it receives the event. The secondary buffer is basically unbounded for us.

index.html
```diff
@@ -854,23 +862,49 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
+<a>resource timing buffer size limit</a> and <a>secondary entry buffer</a> is
+not empty, run the following substeps:
+<ol>
+<li>Let <i>entry</i> be the oldest <a>PerformanceResourceTiming</a> in
+<a>secondary entry buffer</a>.</li>
+<li>Add <i>entry</i> to <a
+data-cite='!PERFORMANCE-TIMELINE-2/#dfn-performance-entry-buffer'>performance
+entry buffer</a>.</li>
+<li>Increment <a>resource timing buffer current size</a> by 1.</li>
+<li>Remove <i>entry</i> from <a>secondary entry buffer</a>.</li>
```
rniwa: Looks like this is missing a step to fire `resourcetimingbufferfull` once the primary buffer becomes full again?

rniwa:
See
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/page/Performance.cpp#L178
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/page/Performance.cpp#L211

index.html
npm1: The point of the buffer size limit is to limit the amount of memory used. Having an unbounded secondary buffer makes that limit pointless.

rniwa: Well, if we're delaying the event dispatching to the next task, how are author scripts supposed to know to clear the buffer? This unboundedness only occurs until the next task, and I don't see any reason why it would be bad. The script authors can just allocate a bunch of arrays to do the same if any.

yoavweiss: While I agree that the unboundedness here would be short lived, I prefer to avoid scenarios where the event is called again on every microtask. So I prefer to define the buffer as bounded to the same size as the RT buffer to guaranty that if the developer either cleared the buffer or doubled its size, all entries will be successfully copied.

This can result in dropped entries when the buffer is too small, but I doubt that would happen in practice, and it makes the processing model simpler.

index.html
yoavweiss: because the secondary buffer is bounded, and we assume developers do the right thing (either clear or double the buffer size), there's no need to fire the event again from here.

yoavweiss: Note that one behavior change with this PR is that the event doesn't fire when the buffer is full, but fired when it *overflows*.

yoavweiss:
> See
> https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/page/Performance.cpp#L178
> https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/page/Performance.cpp#L211

@rniwa I understand that the spec PR diverges from the current WebKit model. However, it seems like a reasonable compromise that enables capping the secondary buffer size without too much complexity.

yoavweiss:
Tests for this PR as well as Chromium implementation are at https://chromium-review.googlesource.com/c/chromium/src/+/1200863

index.html
yoavweiss: > Well, if we're delaying the event dispatching to the next task, how are author scripts supposed to know to clear the buffer?

I'm not sure I understand the issue.
The proposed model is:
* An entry which overflows the buffer, triggers the event queueing.
* The entry as well as following ones go into the secondary buffer.
* Once the event is fired, developers should make room in the primary buffer (either clear or bump its size)
* Right after that, entries are copied from secondary to primary if possible

Do you see a problem with that model?

npm1: @rniwa Just removing this check would cause the second buffer to be filled up indefinitely in the case when ResourceTiming is not used at all. We would need a different check.

@yoavweiss I think the problem being addressed is that potentially the secondary buffer could fill up before the bufferfull event is fired. This would cause the developer to lose some entries with nothing they can do about it.

yoavweiss: I agree that losing entries is a theoretical risk here. I doubt this is something that will happen in real life, assuming developers haven't shrunk their buffer sizes to be unreasonably small.

rniwa: The problem is that the event is dispatched asynchronously. If some script or the HTML/CSS/JS parser ends up requesting more than 150 requests at once and they're cached or failed instantly, then the script may never have a chance to see those entries.

It's not true that this secondary buffer would indefinitely grow when resource timing is never used. If the script never reacts to the event, then we would drop all entries in the secondary buffer after dispatching the very first event.

index.html
rniwa: Okay, that needs to be changed once the secondary buffer's limit is lifted.

rniwa:
FYI, we're not going to change our behavior. Letting the secondary buffer overflow without ever letting the script react to it is simply not acceptable.

index.html
npm1: I'm looking at this PR, not at WebKit's implementation. If you want an unbounded secondary buffer then you cannot just clear this buffer after firing a buffer full event. So yes, it would grow indefinitely.

If we go with an unbounded secondary buffer then I would suggest clearing it only if resource timing buffer current size is equal to resource timing buffer size limit, right after firing the resourcetimingbufferfull event (because this is a strong hint that the entries are not being used).

index.html
npm1: Even if we intend to keep an unbounded secondary buffer, for consistency we might want to fire the additional events async. I see no reason why this one would need be sync. So essentially we would just call the "add a PerformanceResourceTiming entry" algorithm.

rniwa: Making those secondary events async would be fine with me.

index.html
rniwa: You're just picking on my words at that point. Clearly, you have to take my feedback as a whole. What we're proposing is to keep adding entries to the secondary buffer (we should really call it the backup buffer to be consistent with the backup queue we have for custom elements), and then the first time we fire an event on it, if the primary buffer size didn't shrink or the limit hadn't increased, then we would drop all remaining entries in the backup buffer. There won't be no indefinite growth of the backup buffer beyond the first event firing at that point.

npm1: It wasn't clear to me how you proposed getting rid of secondary buffer entries when they were unused. Maybe I missed something? But I think that what you said just now matches what I proposed (looking at the primary buffer size right after firing the event). SGTM

yoavweiss: So, if we have a scenario where `X` entries are accumulated every Task, and on every call to `resourcetimingbufferfull` the developer is increasing the buffer by `Y`, where `Y < X`, this scheme will end up consuming an infinite amount of memory.

That seems like a fairly large risk. The trade-off here, where if the developer didn't dedicate enough space for the entries, they get lost, seems better.

The only problematic scenario is if the initial buffer is significantly smaller then the rate of incoming entries per task. That doesn't seem like a likely scenario with the current browser defaults.

rniwa: Oh now I remember why non-first `resourcetimingbufferfull` events are fired synchronously. This is so that we can ensure that either the buffer is emptied after a single task, or that the remaining entries are dropped on the floor.

Not that if we end up enqueuing more `ResourceTiming` entries as a result of dispatching `resourcetimingbufferfull`, we can fall into an infinite loop here but that's no worse than the current status quo, and this is no different from someone causing a mutual recursion between event listeners.

yoavweiss: What is the benefit of adding all that complexity and edge cases? Why is it all of the sudden OK to fire the events synchronously the second time, but not the first time?

I think a simpler solution would be to basically tell people:
* Don't set the buffer size to ridiculously small amounts. I doubt many people *decrease* the buffer sizes from their default size.
* Clear or significantly increase (at least double) the buffer when it's full

I can see an argument in favor of adding console messages when this happens, so that developers can catch such scenarios, if they at all exist. But adding a ton of complexity in order to save a few entries in theory seem excessive.

rniwa: > Why is it all of the sudden OK to fire the events synchronously the second time, but not the first time?

Because we're already in a new task. There is no security or implementation issues from firing an event synchronously.

yoavweiss: > Because we're already in a new task

new task relative to what?


> There is no security or implementation issues from firing an event synchronously.

Can you expand on why you objected to the origin sync event firing that was in the spec? What were the issues there?

rniwa: The problem is this. There is a lot of code in the browser engine where the resource load gets canceled (e.g. because the document is getting removed from a frame), and adding a resource timing entry in those timing is not permissible.

However, once this new timer fires, there is nothing like that happening down in the stack frame. So it's totally okay to synchronously fire events.

yoavweiss:
As discussed at TPAC, this approach is abandoned. Closing.

PR #168, opened 2018-10-10
Fire `resourcetimingbufferfull` asynchronously - alternative approach 

### LINKED ISSUE #141: When is resourcetimingbufferfull supposed to fire?

yoavweiss:
This is an alternative approach to https://github.com/w3c/resource-timing/pull/163 in order to resolve #141 and #96.

### Handwavy processing

* The secondary buffer is unlimited.
* The event is fired async once, and if the buffer size exceeds its limits after the event fired, it's continuously fired synchronously until the size does not exceed.
* If the event handler does not reduce the buffer size or increases the limit, all entries that exceed the buffer max size get dropped.

/cc @rniwa @tdresser @npm1


igrigorik:
@yoavweiss there is a merge conflict in the PR, could you take a quick look?

@rniwa @tdresser @npm1 friendly bump :)

index.html
```diff
@@ -854,23 +862,64 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
+<ol>
+<li>Let <i>excess entries before</i> be the <a>resource timing buffer current size</a> 
+minus <a>resource timing buffer size limit</a>.</li>
+<li><a data-cite="!DOM/#concept-event-fire">Fire an event</a>
+named <code>resourcetimingbufferfull</code> at the <a data-cite=
+"!HR-TIME-2/#idl-def-performance">Performance</a> object, with its
+<code>bubbles</code> attribute initialized to true.</li>
+<li>Let <i>excess entries after</i> be the <a>resource timing buffer current size</a> 
+minus <a>resource timing buffer size limit</a>.</li>
+<li>If <i>excess entries before</i> equals <i>excess entries after</i>, then while <a>resource timing buffer current size</a> is larger than <a>resource timing buffer size limit</a>, run the following substeps:
```
rniwa: This isn't quite right. There is no guarantee that the secondary buffer is empty at this point.
We need to start dropping things from the secondary buffer first.

yoavweiss: In step 2 of [fire the buffer full event](https://pr-preview.s3.amazonaws.com/yoavweiss/resource-timing/pull/168.html#dfn-fire-a-buffer-full-event) we clear out the secondary buffer. Then if the number of entries exceeds the buffer size, we fire the event synchronously, and if the buffer size remained the same, we drop entries.

How could the secondary buffer be non-empty at this point? Are you suggesting that "add a PerformanceResourceTiming entry" call happened while these sync steps are running?

rniwa: As I mentioned during F2F, firing of `resourcetimingbufferfull` event itself can result in more resource timing entries to be added; e.g. via a sync XHR.

npm1:
We decided to go with this alternative, right? Is this ready for review now?

yoavweiss:
> We decided to go with this alternative, right? Is this ready for review now?

Not yet, as @rniwa's comments are still pending. Looking at that now

yoavweiss:
> @yoavweiss there is a merge conflict in the PR, could you take a quick look?

Missed that comment... Rebasing now

index.html
yoavweiss: @rniwa - Fixed the processing to fire the copying from the secondary buffer after the sync event firing (and deleting everything that exceeds if the buffer is equal or larger).
PTAL

yoavweiss:
> We decided to go with this alternative, right? Is this ready for review now?

Ready for review now :)

yoavweiss:
Had to manually rebase and force push due to merge shenanigans. Should be stable now...

index.html
```diff
@@ -833,23 +839,70 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-"HR-TIME-2#idl-def-performance">Performance</a> object, with its
-<code>bubbles</code> attribute initialized to true.</li>
+<ol style="list-style-type: lower-latin;">
+<li>Set <a>resource timing buffer full flag</a> to true.</li>
+<li>Queue a task to run <a>fire a buffer full event</a>.</li>
+</ol>
+<li>Add <i>new entry</i> to the <a>secondary entry buffer</a>.</li>
+</ol>
+
+<p>To <dfn>copy secondary buffer</dfn>, run the following steps:</p>
```
npm1: I think this algorithm needs to check that the performance entry buffer size limit is never exceeded.

index.html
```diff
@@ -833,23 +839,70 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
+<p>To <dfn>fire a buffer full event</dfn>, run the following steps:</p>
+<ol>
+<li><a data-cite="DOM/#concept-event-fire">Fire an event</a>
+named <code>resourcetimingbufferfull</code> at the <a data-cite=
+"hr-time-2/#idl-def-performance">Performance</a> object,</li>
+<li>Run <a>copy secondary buffer</a>.</li>
+<li>While <a>resource timing buffer current size</a> is larger than 
+<a>resource timing buffer size limit</a>, run the following substeps:
+<ol>
+<li>Let <i>excess entries before</i> be the <a>resource timing buffer current
```
npm1: `number of excess entries before`? Maybe it's too long but from this name it's clear that it's the number not the actual entries.

index.html
npm1: To clarify, the reason this is important is that developers could query entries in the buffer full event, and this algorithm is saying that they will get all of them even though they're not supposed to be in the performance timeline yet.

index.html
```diff
@@ -833,23 +839,70 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
+<li>Assert that <a>secondary entry buffer</a> is empty.</li>
+</ol>
+
+<p>To <dfn>fire a buffer full event</dfn>, run the following steps:</p>
+<ol>
+<li><a data-cite="DOM/#concept-event-fire">Fire an event</a>
+named <code>resourcetimingbufferfull</code> at the <a data-cite=
+"hr-time-2/#idl-def-performance">Performance</a> object,</li>
+<li>Run <a>copy secondary buffer</a>.</li>
+<li>While <a>resource timing buffer current size</a> is larger than
```
npm1: And then this while asks about emptiness of secondary buffer instead of primary buffer size.

index.html
```diff
@@ -833,23 +839,70 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
+buffer</a>.</li>
+<li>Increment <a>resource timing buffer current size</a> by 1.</li>
+<li>Remove <i>entry</i> from <a>secondary entry buffer</a>.</li>
+</ol>
+<li>Assert that <a>secondary entry buffer</a> is empty.</li>
+</ol>
+
+<p>To <dfn>fire a buffer full event</dfn>, run the following steps:</p>
+<ol>
+<li><a data-cite="DOM/#concept-event-fire">Fire an event</a>
```
npm1: Can this be introduced into the while?

index.html
yoavweiss: so instead of sync firing the event when the primary exceeds, we'd fire it until the secondary is empty? I guess that can work

index.html
yoavweiss: ok

index.html
yoavweiss: yeah

index.html
yoavweiss: yup

index.html
```diff
@@ -786,6 +786,8 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
 initially 0.</li>
 <li>A <dfn>resource timing buffer full flag</dfn> which is
 initially false.</li>
+<li>A <dfn>resource timing secondary buffer current size</dfn> which is
+initially 0.</li>
 <li>A <dfn>secondary entry buffer</dfn> to store <a>PerformanceEntry</a>
```
npm1: nit: inconsistent dfns. One includes 'resource timing', like the rest, the other does not.

npm1: Also inconsistency between 'secondary buffer' and 'secondary entry buffer'

index.html
```diff
@@ -862,35 +867,28 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
 
 <p>To <dfn>fire a buffer full event</dfn>, run the following steps:</p>
 <ol>
-<li><a data-cite="DOM/#concept-event-fire">Fire an event</a>
-named <code>resourcetimingbufferfull</code> at the <a data-cite=
-"hr-time-2/#idl-def-performance">Performance</a> object,</li>
-<li>Run <a>copy secondary buffer</a>.</li>
-<li>While <a>resource timing buffer current size</a> is larger than 
-<a>resource timing buffer size limit</a>, run the following substeps:
+<li>While <a>resource timing buffer current size</a> equals
```
npm1: This indicates that `resourcetimingbufferfull` will not be fired if the buffer is cleared between the time we queue the task and the time we run the task. This seems reasonable to me.

index.html
```diff
@@ -817,14 +823,16 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-<p>To <dfn>add a PerformanceResourceTiming entry</dfn> (<i>new
-entry</i>) in the <a data-cite=
-'PERFORMANCE-TIMELINE-2#dfn-performance-entry-buffer'>performance
+
+<p>To <dfn>add a PerformanceResourceTiming entry</dfn> into the <a data-cite=
+'PERFORMANCE-TIMELINE-2/#dfn-performance-entry-buffer'>performance
 entry buffer</a>, run the following steps:</p>
 <ol>
+<li>Let <i>new entry</i> be the input <a>PerformanceEntry</a> to be added.
 <li>If <a>resource timing buffer current size</a> is less than
```
npm1: Hmm I think we also need to check that the secondary buffer is empty here? Otherwise we're adding entries in the incorrect order. That is, if we get full, queue a task, and add some entries into secondary buffer but then clear the primary buffer and then get a new entry, we'll add that new entry to the primary buffer before the ones in the secondary buffer.

index.html
npm1: Oh but actually there does not seem to be a guarantee that secondary buffer will be empty by the end of this algorithm? I think this task should be roughly as follows:
`While primary full OR secondary buffer not empty:`
`if primary full then fire resourcetimingbufferfull`
`if secondary nonempty then copy secondary. if nothing copied, empty secondary.`

index.html
yoavweiss: ok

index.html
yoavweiss: Yeah, we need to check the `resource timing buffer full flag` before adding to the primary buffer

index.html
yoavweiss: Good catch.

yoavweiss: Well, trying it out, the above algorithm ends up with an infinite loop, as the primary can be full but with no excess entries, in which case, no event should fire. Will fix.

yoavweiss: With the tests to keep me honest, I ended up with something like: "While secondary buffer is not empty: if primary full, then fire resourcetimingbufferfull; Copy secondary; If secondary is same size as before, empty secondary and abort"

yoavweiss: To clarify: The difference between "if nothing copied" and "same size as before" is for cases where resources where copied, but the same amount of resources was added inside the call. "if nothing copied" opens up a potential for infinite loops when the handler increases by one but also triggers a sync resource call at the same time.

index.html
```diff
@@ -786,10 +786,10 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
 <li>A <dfn>resource timing buffer full flag</dfn> which is
 initially false.</li>
-<li>A <dfn>resource timing secondary buffer current size</dfn> which is
-initially 0.</li>
-<li>A <dfn>secondary entry buffer</dfn> to store <a>PerformanceEntry</a>
-objects that is initially empty.</li>
+<li>A <dfn>resource timing secondary buffer current size</dfn>
+which is initially 0.</li>
+<li>A <dfn>resource timing secondary buffer</dfn> to store
+<a>PerformanceEntry</a> objects that is initially empty.</li>
```
npm1: nit: it stores PerformanceResourceTiming objects in particular.

yoavweiss: true

index.html
```diff
@@ -828,14 +828,21 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
+
 <p>To <dfn>add a PerformanceResourceTiming entry</dfn> into the <a data-cite=
 'PERFORMANCE-TIMELINE-2/#dfn-performance-entry-buffer'>performance
 entry buffer</a>, run the following steps:</p>
 <ol>
 <li>Let <i>new entry</i> be the input <a>PerformanceEntry</a> to be added.
-<li>If <a>resource timing buffer current size</a> is less than
-<a>resource timing buffer size limit</a>, run the following
-substeps:
+<li>If <a>can add resource timing entry</a> returns true and <a>resource
```
npm1: This does not solve the problem of the example I mentioned earlier:

- Buffer becomes full and we queue the buffer full task.
- We receive more entries, so they are added to the secondary buffer.
- We clear the primary buffer, thus the buffer full flag becomes false. Secondary buffer still has entries.
- We receive a new entry.

Per the current algorithm, this new entry would be added to the primary buffer before the ones in the secondary have had a chance to go into the primary buffer. So, we should check secondary buffer. If it has entries, then add the new entry to the secondary buffer.

index.html
npm1: Ok looks much better! However, I still think "nothing copied" is slightly better here. If nothing copied, we can empty the secondary buffer and abort. Since the while now only depends on secondary buffer being nonempty, that will not cause infinite loops.

yoavweiss: It will if every event that fires adds an entry to the secondary buffer.  That way, the sync event loop will continue forever, because every iteration copies a single entry. Don't know how realistic that scenario is, but overall doesn't seem worth the risk.

index.html
yoavweiss: OK, changed the semantics of the "buffer full flag" to be "buffer full event pending flag", which resolves this issue. After discussing with @npm1, they managed to find a scenario where this is actually observable to make it testable. Should be all good now.

yoavweiss:
Tests for this are being reviewed at https://chromium-review.googlesource.com/c/chromium/src/+/1345269
I'll land both at the same time.

mstange:
Firefox issue is https://bugzilla.mozilla.org/show_bug.cgi?id=1510374

PR #169, opened 2018-10-10
Add a note about "stalled" or "blocked" time

### LINKED ISSUE #164: elaborate on `stalled` or `blocked` stage in specs
Devtools of modern browsers all got this handy waterfall graph for each resource timing, although well explained in each vendors' documentation ([google](https://developers.google.com/web/tools/chrome-devtools/network-performance/reference#timing-explanation), [firefox](https://developer.mozilla.org/en-US/docs/Tools/Network_Monitor#Timings)), I think it will help more people to get the gist of how resource is loaded if we add some non-normative section in the spec to elaborate on the `stalled` or `queued` duration.

If this seems to be feasible, I'd like to send a pull request regarding this issue.

image
screenshot of resource timing graph on Google Chrome 

image
screenshot of resource timing graph on Mozilla Firefox

jasonslyvia:
Fixes #164

***
<a href="https://pr-preview.s3.amazonaws.com/jasonslyvia/resource-timing/pull/169.html" title="Last updated on Oct 10, 2018, 3:48 PM GMT (2e660ed)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/169/de1f8b5...jasonslyvia:2e660ed.html" title="Last updated on Oct 10, 2018, 3:48 PM GMT (2e660ed)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -947,6 +947,12 @@ does not pass the <a>timing allow check</a> algorithm.</figcaption>
 <!-- Source: https://docs.google.com/document/d/1I7XGNJ57Qgjkg9pL11s7MK7zGEcwAgdNj1W5f7NKbW8/ -->
 <img src="timestamp-diagram.svg" alt="Resource Timing attributes"
 style='margin-top: 1em'></figure>
+<aside class="note">
+<p>The phases illustrated above not necessarily occur one after another,
+there may be time spent waiting for a connection to be available mostly due
+to concurrent TCP connections limit of HTTP/1.X, which is commonly known as
+"stalled" or "blocked" time.</p>
+</aside>
 <p>For each resource <a data-cite=
 "!FETCH/#concept-fetch">fetched</a> by the current <a data-cite=
 "!HTML/#browsing-context">browsing context</a>, excluding resources


index.html
```diff
@@ -947,6 +947,12 @@ <h3>Processing Model</h3>
 <!-- Source: https://docs.google.com/document/d/1I7XGNJ57Qgjkg9pL11s7MK7zGEcwAgdNj1W5f7NKbW8/ -->
 <img src="timestamp-diagram.svg" alt="Resource Timing attributes"
 style='margin-top: 1em'></figure>
+<aside class="note">
```
yoavweiss: Maybe "The phases illustrated above are not necessarily consecutive, and user agents may spend time between them on processing, waiting for TCP connections to become available, waiting before sending requests to prevent contention or any other implementation detail. The unaccounted time spent by the user agent can be calculated as the difference between the consecutive timestamps. Its sum can be calculated as the difference between the `duration` value and `responseEnd - startTime`. "

@nicjansma @igrigorik - thoughts?

igrigorik: Hmm, tbh I find this really cryptic and hard to grok. As in, you have to be well versed in the nuance here to understand what this is saying. If our concern is specifically "stalled" / "blocked", I'd focus and lead with that.

"The user agent may inject other processing steps that are not included in the above diagram. For example, due to limitations of HTTP/1.x TCP connection limits, the user agent may be blocked waiting for an available socket, and such delays are not explicitly exposed by this specification."

~ish?

jasonslyvia: As per the fact that many developers collect performance metrics based on this diagram, personally I prefer @yoavweiss 's version since it presents a clear formula for calculating "stalled" time

mmocny:
Closing, since https://github.com/w3c/resource-timing/issues/164 is already fixed via other means.

PR #170, opened 2018-10-11
editorial: update links

marcoscaceres:
Fixup linking - remove some badness (e.g., reference to WebIDL-1 and HTML5).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/pull/170.html" title="Last updated on Oct 11, 2018, 7:11 AM GMT (b2a5f02)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/170/de1f8b5...b2a5f02.html" title="Last updated on Oct 11, 2018, 7:11 AM GMT (b2a5f02)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
 <meta name="generator" content=
-"HTML Tidy for HTML5 for Mac OS X version 5.5.31">
+"HTML Tidy for HTML5 for Apple macOS version 5.6.0">
 <meta charset="utf-8">
 <title>Resource Timing Level 2</title>
 
@@ -52,17 +52,7 @@ async class='remove'></script>
     wgPublicList: "public-web-perf",
     subjectPrefix: "[ResourceTiming]",
     github: "https://github.com/w3c/resource-timing/",
-    otherLinks: [
-      {
-        key: 'Implementation',
-        data: [
-          {
-            value: 'Can I use Resource Timing?',
-            href: 'http://caniuse.com/#feat=resource-timing'
-          }
-        ],
-      }
-    ],
+    caniuse: "resource-timing",
     wgPatentURI:  "https://www.w3.org/2004/01/pp-impl/45211/status"
   };
 </script>
@@ -181,10 +171,6 @@ may be implemented in any manner, so long as the end result is
 equivalent. (In particular, the algorithms defined in this
 specification are intended to be easy to follow, and not intended
 to be performant.)</p>
-<p>The IDL fragments in this specification must be interpreted as
-required for <a data-cite=
-"!WEBIDL-1/#dfn-conforming-set-of-idl-fragments">conforming IDL
-fragments</a>, as described in the [[!WEBIDL-1]] specification.</p>
 </section>
 <section id="terminology">
 <h2>Terminology</h2>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
The HTML links might be broken now... but please don't point to the outdated "HTML5".

marcoscaceres:
Note: when you run tidy, you should run it like this: 
`tidy -config tidyconfig.txt -o index.html index.html`

index.html
```diff
@@ -237,36 +223,36 @@ <h3>Introduction</h3>
...
 Timeline</a> of the relevant context. Resources that are retrieved
-from <a data-cite="!HTML/#relevant-application-cache">relevant
+from <a data-cite="HTML#relevant-application-cache">relevant
 application caches</a> or local resources MUST be included as
 <a>PerformanceResourceTiming</a> objects in the <a data-cite=
-"!PERFORMANCE-TIMELINE-2/#performance-timeline">Performance
-Timeline</a> [[!PERFORMANCE-TIMELINE-2]]. Resources for which the
-<a data-cite="!FETCH/#concept-fetch">fetch</a> was initiated, but
-was later aborted (e.g. due to a network error) MAY be included as
+"PERFORMANCE-TIMELINE-2#performance-timeline">Performance
```
yoavweiss: While we're at it, is it possible to point to the WD, rather than the TR? (Potentially at a different PR)

yoavweiss: To be clear, I can do that, just asking if it's possible (and if so, what's the magic spec name)

marcoscaceres: Yes, but you need to check the key on https://www.specref.org and add it if missing.

index.html
```diff
@@ -813,13 +788,13 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
 <p>The <dfn>Performance</dfn> interface is defined in
-[[!HR-TIME-2]].</p>
+[[HR-TIME-2]].</p>
 <p>The method <dfn>clearResourceTimings</dfn> runs the following
 steps:</p>
 <ol>
 <li>remove all <a>PerformanceResourceTiming</a> objects in the
 <a data-cite=
-'!PERFORMANCE-TIMELINE-2/#dfn-performance-entry-buffer'>performance
+'!PERFORMANCE-TIMELINE-2#dfn-performance-entry-buffer'>performance
```
yoavweiss: So here we need the "!" prefix? If so why? WHY???????

yoavweiss: Also, can you point me to relevant docs that'd allow me to educate myself on that front?

index.html
yoavweiss: Looking at specref, it seems like the ED URL is already there in the w3c.json, but as "edDraft" rather then the "href". Is it possible to point to that URL from the caller? Or do I need to add a separate entry where the "href" is the ED draft URL?

index.html
marcoscaceres: Oh, you missed the memo 📝 https://lists.w3.org/Archives/Public/spec-prod/2018OctDec/0000.html

We now use “artificial intelligence” (i.e., “if statements”) for citations, so ReSpec knows if a reference is normative or informative:)

marcoscaceres: So, we don’t actually need “!” anywhere, my mistake when I did find/replace.

yoavweiss: I'm glad our robot overlords are being kind with us. Can you remove the '!' then?

marcoscaceres: Yeah, remove it. Sorry, already left the office, so on my phone... otherwise would do it now.

yoavweiss: will do!

Issue #152, opened 2018-04-14
`origin-or-null` definition and TAO processing


yoavweiss:
I'm failing to find if and where `origin-or-null` is defined. The "null" case is also not addressed in the "timing allow check" processing model.

IIUC, we need to:
* Add a definition for `origin-or-null`. I'm not sure if the `origin` part should be a token or there are more specific syntax definitions at https://tools.ietf.org/html/rfc7230#appendix-B or elsewhere (CCing @reschke for his opinion)
* Point the origin comparison in the processing model to https://tools.ietf.org/html/rfc6454#section-5 as https://tools.ietf.org/html/rfc6454#section-4 doesn't really return a string.
* Add tests that make sure that origin comparison in TAO also works for non-standard ports, doesn't work for different schemes, etc.
* Add a `null` related condition in the processing model.

Thoughts?

igrigorik:
`origin-or-null` is defined in Fetch: https://fetch.spec.whatwg.org/#origin-header

yoavweiss:
OK, cool. So we can just link to that definition. The rest of the points still stand.

yoavweiss:
Following a [discussion on IRC](https://freenode.logbot.info/whatwg/20181012#c1763402) with @annevk, I tend to think this will require larger Fetch integration than I initially thought:
* "null" handling should work for tainted or opaque origins
* In order to support that, we're better off relying on some version of [Fetch's serializing a request origin](https://fetch.spec.whatwg.org/#serializing-a-request-origin), only we'll need an equivalent using the Document's origin. I'm not quite sure how that would work.
* We'd need tests for all of the above.

annevk:
It seems that a number of these things you can and should address today, either by fixing or adding issue markers, so implementers don't implement security vulnerabilities.

yoavweiss:
> You'd invoke https://html.spec.whatwg.org/#ascii-serialisation-of-an-origin directly on document's origin (which is defined at https://dom.spec.whatwg.org/#concept-document-origin by the way).

OK, so that's one missing piece that would cover opaque origins and is fairly easy to add.



> The main problem I see with your current setup is that you have no notion of redirects so if A requests B which redirects to A, this would consider it same-origin, despite it leaking information about B. That seems bad.

[...]

PR #172, opened 2018-10-12
Use HTML's origin serialization, add note regarding TAO and redirects

yoavweiss:
Addresses some of the concerns for https://github.com/w3c/resource-timing/issues/152

***
<a href="https://pr-preview.s3.amazonaws.com/yoavweiss/resource-timing/pull/172.html" title="Last updated on Oct 21, 2018, 9:36 AM GMT (23c434b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/172/cd2e3ec...yoavweiss:23c434b.html" title="Last updated on Oct 21, 2018, 9:36 AM GMT (23c434b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -894,8 +894,8 @@ whether a resource's timing information can be shared with the
 <li>
 <p>If the <a>Timing-Allow-Origin</a> header value list contains a
 value which is byte-for-byte identical to the <a
-data-cite="HTML#ascii-serialisation-of-an-origin">ascii serialized value</a>
-of the <a>current document</a>'s <a data-cite="HTML#concept-origin">origin</a>,
+data-cite="HTML#ascii-serialisation-of-an-origin">serialisation</a>
+of the <a>current document</a>'s <a data-cite="DOM#concept-document-origin">origin</a>,
 or a wildcard ("<code>*</code>"), return <code>pass</code>.</p>
 </li>
 <li>Return <code>fail</code>.</li>
--- a/index.html
+++ b/index.html
@@ -900,7 +900,7 @@ or a wildcard ("<code>*</code>"), return <code>pass</code>.</p>
 </li>
 <li>Return <code>fail</code>.</li>
 </ol>
-<p class="issue" data-number="152">Above "timing allow check" algorithm should also verify that 
+<p class="issue" data-number="152">The above <a data-cite="#dfn-timing-allow-check">timing allow check</a> algorithm should also verify that 
 no cross-origin redirections were involved in the response, as that would 
 leak timing information about those origins.
 </p>
--- a/index.html
+++ b/index.html
@@ -893,13 +893,17 @@ whether a resource's timing information can be shared with the
 </li>
 <li>
 <p>If the <a>Timing-Allow-Origin</a> header value list contains a
-case-sensitive match for the value of the <code><a data-cite=
-"RFC6454#section-4" data-lt="http-origin">origin</a></code> of the
-<a>current document</a>, or a wildcard ("<code>*</code>"), return
-<code>pass</code>.</p>
+value which is byte-for-byte identical to the <a
+data-cite="HTML#ascii-serialisation-of-an-origin">ascii serialized value</a>
+of the <a>current document</a>'s <a data-cite="HTML#concept-origin">origin</a>,
+or a wildcard ("<code>*</code>"), return <code>pass</code>.</p>
 </li>
 <li>Return <code>fail</code>.</li>
 </ol>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -893,13 +893,17 @@ <h4><code>Timing-Allow-Origin</code> Response Header</h4>
 </li>
 <li>
 <p>If the <a>Timing-Allow-Origin</a> header value list contains a
-case-sensitive match for the value of the <code><a data-cite=
-"RFC6454#section-4" data-lt="http-origin">origin</a></code> of the
-<a>current document</a>, or a wildcard ("<code>*</code>"), return
-<code>pass</code>.</p>
+value which is byte-for-byte identical to the <a
+data-cite="HTML#ascii-serialisation-of-an-origin">serialisation</a>
```
annevk: serialization*, unless you really use en-GB. (I realize it's confusing that the fragment does use en-GB, but that's hard to change other than keeping two long fragments around forever.)

yoavweiss: I really don't. The laptop I switched to recently keeps trying to force me into GB spelling, which is probably the cause here.

index.html
```diff
@@ -893,13 +893,17 @@ <h4><code>Timing-Allow-Origin</code> Response Header</h4>
...
+data-cite="HTML#ascii-serialisation-of-an-origin">serialization</a>
+of the <a>current document</a>'s <a data-cite="DOM#concept-document-origin">origin</a>,
+or a wildcard ("<code>*</code>"), return <code>pass</code>.</p>
 </li>
 <li>Return <code>fail</code>.</li>
 </ol>
+<p class="issue" data-number="152">Above "timing allow check" algorithm should also verify that 
+no cross-origin redirections were involved in the response, as that would 
+leak timing information about those origins.
+</p>
```
igrigorik: Doesn't it do that already? As in, we run TAO check on every response in the processing model and reset the values if the check fails.

yoavweiss: We don't have the concept of a "tainted origin" until we integrate with Fetch. The current processing model talks about "the last non-redirected fetch of the resource", but as @annevk pointed out, there can be a SameOrigin=>X-Origin=>SameOrigin scenario where the spec as is can expose cross-origin information.

igrigorik: What would it expose in this case?

annevk: The idea is that X-Origin could influence the final SameOrigin in such a way as to attack it. CORS also requires the appropriate headers in such a scenario.

index.html
```diff
@@ -893,13 +893,17 @@ <h4><code>Timing-Allow-Origin</code> Response Header</h4>
...
-<a>current document</a>, or a wildcard ("<code>*</code>"), return
-<code>pass</code>.</p>
+value which is byte-for-byte identical to the <a
+data-cite="HTML#ascii-serialisation-of-an-origin">serialization</a>
+of the <a>current document</a>'s <a data-cite="DOM#concept-document-origin">origin</a>,
+or a wildcard ("<code>*</code>"), return <code>pass</code>.</p>
 </li>
 <li>Return <code>fail</code>.</li>
 </ol>
+<p class="issue" data-number="152">Above "timing allow check" algorithm should also verify that
```
cvazac: s/Above/The above/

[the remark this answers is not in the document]
yoavweiss: yeah, makes sense

index.html
yoavweiss: One scenario could be:
* malicious.com redirects a resource to victim.com (which unfortunately exposes a redirection endpoint, which can redirect back to any arbitrary URL). It uses the redirection endpoint to redirect back to malicious.com.
* With current language, that would expose timing information about victim.com, which may be used to conclude's the user's login state, or other personal information.

annevk: That example doesn't really illustrate why malicious.com would need a TAO header on the final response. This is victim.com requesting something from target.com, which is either compromised or redirects to a a page on victim.com in order to confuse victim.com (or there are even more domains in the chain). And in that case we want victim.com to explicitly opt-in so that we know it agreed to be confused.

yoavweiss: @annevk given they (handwavy) tackling of this in https://github.com/w3c/resource-timing/pull/196, do we still need an explicit note here in your opinion?

annevk: Yes, because the timing allow check itself returns "pass" if the "current resource" is same-origin with the requesting document. That means that EXAMPLE -> ATTACKER -> EXAMPLE would leak because ATTACKER would surely set the header and EXAMPLE does not need to.

yoavweiss:
This PR is obsoleted by https://github.com/w3c/resource-timing/pull/197

PR #177, opened 2018-10-31
Better define which entries are exposed for nested contexts

### LINKED ISSUE #166: Need to define an actual processing model for what things are added to resource timing and for what global
Some examples that are completely undefined right now, as far as I can tell:

1. If an iframe is navigated via changing the src attribute, does that load get added to resource timing?  For which page?

2. If an iframe is navigated via setting its location.href, same questions.

3. If an iframe is navigated via meta refresh, same questions.

Similar questions for an `<object>` or `<embed>`,  with the additional complication that for those you don't know up front whether the load is a navigation or not.

yoavweiss:
Closes #166

### DIFF
--- a/index.html
+++ b/index.html
@@ -927,7 +927,7 @@ does not pass the <a>timing allow check</a> algorithm.</figcaption>
 <img src="timestamp-diagram.svg" alt="Resource Timing attributes"
 style='margin-top: 1em'></figure>
 
-<p>For each resource which <a>Request</a> has a non-null <a>client</a>, perform
+<p>For each resource whose <a>Request</a> has a non-null <a>client</a>, perform
 the following steps:</p>
 <ol data-link-for="PerformanceResourceTiming">
 <li>If the resource is <a data-cite="Fetch#concept-fetch">fetched</a> by a
@@ -941,7 +941,7 @@ surface resource fetch events.</p>
 <div class="issue" data-number="27"></div>
 <li>If the resource's <a>Request</a>'s
 <a data-cite="Fetch#concept-request-destination">destination</a> equals to
-"document", and was not triggered by 
+"document", and the <a>Request</a> was not triggered by 
 <a data-cite="HTML#process-the-iframe-attributes">process the iframe attributes</a> or
 <a data-cite="HTML#process-the-frame-attributes">process the frame attributes</a>,
 abort the remaining steps.</li>
--- a/index.html
+++ b/index.html
@@ -238,13 +238,18 @@ elements [[SVG11]] such as <a data-cite=
 <section>
 <h3>Resources Included in the <a>PerformanceResourceTiming</a>
 Interface</h3>
-<p>All resources <a data-cite="FETCH#concept-fetch">fetched</a> by
-the current <a data-cite="HTML#browsing-context">browsing</a> or
-worker [[WORKERS]] context's MUST be included as
+<p>All resource <a data-cite="Fetch#concept-request">Request</a>s
+<a data-cite="FETCH#concept-fetch">fetched</a> by the current
+<a data-cite="HTML#browsing-context">browsing context</a>'s
+<a data-cite="HTML#active-document">active document</a>'s
+<a data-cite="HTML#relevant-settings-object">relevent settings object</a>
+MUST be included as
 <a>PerformanceResourceTiming</a> objects in the <a data-cite=
 "PERFORMANCE-TIMELINE-2#performance-timeline">Performance
-Timeline</a> of the relevant context. Resources that are retrieved
-from <a data-cite="HTML#relevant-application-cache">relevant
+Timeline</a> of the
+<a data-cite="HTML#concept-relevant-global">relevant global object</a>.
+Resources that are retrieved from 
+<a data-cite="HTML#relevant-application-cache">relevant
 application caches</a> or local resources MUST be included as
 <a>PerformanceResourceTiming</a> objects in the <a data-cite=
 "PERFORMANCE-TIMELINE-2#performance-timeline">Performance
@@ -922,9 +927,13 @@ does not pass the <a>timing allow check</a> algorithm.</figcaption>
 <!-- Source: https://docs.google.com/document/d/1I7XGNJ57Qgjkg9pL11s7MK7zGEcwAgdNj1W5f7NKbW8/ -->
 <img src="timestamp-diagram.svg" alt="Resource Timing attributes"
 style='margin-top: 1em'></figure>
-<p>For each resource <a data-cite="FETCH#concept-fetch">fetched</a>
-by the current <a data-cite="HTML#browsing-context">browsing
-context</a>, excluding resources fetched by cross-origin
+
+<p>For each resource which <a data-cite="Fetch#concept-request">Request</a> is
+<a data-cite="FETCH#concept-fetch">fetched</a> by the current
+<a data-cite="HTML#browsing-context">browsing context</a>'s
+<a data-cite="HTML#active-document">active document</a>'s
+<a data-cite="HTML#relevant-settings-object">relevent settings object</a>,
+excluding resources fetched by cross-origin
 stylesheets fetched with <code>no-cors</code> policy, perform the
 following steps:</p>
 <p class="issue">Above cross-origin exclusion should be defined via
@@ -934,6 +943,13 @@ turn, Resource Timing should interface with Fetch registry to
 surface resource fetch events.</p>
 <div class="issue" data-number="27"></div>
 <ol data-link-for="PerformanceResourceTiming">
+<li>If the resource's <a data-cite="Fetch#concept-request">Request</a>'s
+<a data-cite="Fetch#concept-request-destination">destination</a> equals to
+"document", "object" or "embed", and was not triggered by 
+<a data-cite="HTML#process-the-iframe-attributes">process the iframe attributes</a>,
+<a data-cite="HTML#process-the-object-attributes">process the object attributes</a>, or
+<a data-cite="HTML#process-the-embed-attributes">process the embed attributes</a>, 
+respectively, abort the remaining steps.</li>
 <li><dfn data-lt="step-create-object">Create a new
 <a>PerformanceResourceTiming</a> object</dfn> and set
 <a>entryType</a> to the DOMString <code>resource</code>.</li>
--- a/index.html
+++ b/index.html
@@ -238,16 +238,15 @@ elements [[SVG11]] such as <a data-cite=
 <section>
 <h3>Resources Included in the <a>PerformanceResourceTiming</a>
 Interface</h3>
-<p>All resource <a data-cite="Fetch#concept-request">Request</a>s
-<a data-cite="FETCH#concept-fetch">fetched</a> by the current
-<a data-cite="HTML#browsing-context">browsing context</a>'s
-<a data-cite="HTML#active-document">active document</a>'s
-<a data-cite="HTML#relevant-settings-object">relevent settings object</a>
-MUST be included as
-<a>PerformanceResourceTiming</a> objects in the <a data-cite=
+<p>All resource <dfn data-cite="Fetch#concept-request">Request</dfn>s
+<a data-cite="FETCH#concept-fetch">fetched</a> by a non-null <dfn
+data-cite="Fetch#concept-request-client">client</dfn> MUST be included as
+<a>PerformanceResourceTiming</a> objects in the <a>client</a>'s <dfn
+data-cite="HTML#concept-settings-object-global">global object</dfn>'s
+<a data-cite=
 "PERFORMANCE-TIMELINE-2#performance-timeline">Performance
-Timeline</a> of the
-<a data-cite="HTML#concept-relevant-global">relevant global object</a>.
+Timeline</a>, unless excluded from the timeline as part of the <a
+href="#processing-model">processing model</a>.
 Resources that are retrieved from 
 <a data-cite="HTML#relevant-application-cache">relevant
 application caches</a> or local resources MUST be included as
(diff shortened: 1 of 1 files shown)


yoavweiss:
cc @bzbarsky - FYI. Would be great if you could review as well

yoavweiss:
Related tests are at https://github.com/web-platform-tests/wpt/pull/13823

index.html
```diff
@@ -238,13 +238,18 @@ <h3>Introduction</h3>
...
 <h3>Resources Included in the <a>PerformanceResourceTiming</a>
 Interface</h3>
-<p>All resources <a data-cite="FETCH#concept-fetch">fetched</a> by
-the current <a data-cite="HTML#browsing-context">browsing</a> or
-worker [[WORKERS]] context's MUST be included as
+<p>All resource <a data-cite="Fetch#concept-request">Request</a>s
+<a data-cite="FETCH#concept-fetch">fetched</a> by the current
+<a data-cite="HTML#browsing-context">browsing context</a>'s
+<a data-cite="HTML#active-document">active document</a>'s
+<a data-cite="HTML#relevant-settings-object">relevent settings object</a>
```
annevk: You dropped the bit about workers. It's also rather hard to review this as it's unclear what the context for "current" is. The user agent? If you're already in some scope, this seems wrong.

index.html
```diff
@@ -934,6 +943,13 @@ <h3>Processing Model</h3>
 surface resource fetch events.</p>
 <div class="issue" data-number="27"></div>
 <ol data-link-for="PerformanceResourceTiming">
+<li>If the resource's <a data-cite="Fetch#concept-request">Request</a>'s
+<a data-cite="Fetch#concept-request-destination">destination</a> equals to
+"document", "object" or "embed", and was not triggered by 
+<a data-cite="HTML#process-the-iframe-attributes">process the iframe attributes</a>,
+<a data-cite="HTML#process-the-object-attributes">process the object attributes</a>, or
+<a data-cite="HTML#process-the-embed-attributes">process the embed attributes</a>,
```
annevk: Hmm, I don't think iframe should be special here.

index.html
yoavweiss: What would be a better way to define "current"? You're right about workers. How can I better refer them here?

index.html
yoavweiss: The reasoning behind it is that the top level document should have visibility to iframe loads it initiated (or that were initiated in its context), but should not have visibility into iframe loads initiated by the iframe's content. I'm open to other ways to define the above.

/cc @bzbarsky

index.html
annevk: Where in the spec is this used? If this is some API it ideally would return some pre-existing collection that other specifications produce.

index.html
annevk: Oh, that seems somewhat of a mismatch with how fetch groups should work. Also, what if I set `contentWindow.location`, I guess that doesn't count?

index.html
```diff
@@ -238,13 +238,18 @@ <h3>Introduction</h3>
 <section>
 <h3>Resources Included in the <a>PerformanceResourceTiming</a>
 Interface</h3>
-<p>All resources <a data-cite="FETCH#concept-fetch">fetched</a> by
-the current <a data-cite="HTML#browsing-context">browsing</a> or
-worker [[WORKERS]] context's MUST be included as
+<p>All resource <a data-cite="Fetch#concept-request">Request</a>s
+<a data-cite="FETCH#concept-fetch">fetched</a> by the current
```
bzbarsky: What is the "current browsing context"?  That's a longstanding issue here...  The PerformanceResourceTiming is tied to a global and hence settings object.  That's the settings object that should presumably be relevant here?

One option might be that given a fetch we get its [**client**](https://fetch.spec.whatwg.org/#concept-request-client), if not null get the PerformanceResourceTiming for that settings object's global (need to word this carefully around workers) and use that.  I _think_ that will do the right thing for iframes, because loads due to attr changes use the iframe's nodee document's window as the client, which is what we want here.

Doing this on the browsing context level, if actually implemented that way (the way the spec is trying to define it here), would be a security hole: a PerformanceResourceTiming from an earlier page in the history would be required to see loads from later pages in the history.   That's clearly not desirable.

index.html
bzbarsky: Right, the intent is to exclude navigations due to `location` sets, link clicks in the iframe, etc, etc.

I'm not 100% sure how fetch groups should relate to this.  Which fetch group is the load for an `<iframe src="stuff">` supposed to be in?

index.html
```diff
@@ -934,6 +943,13 @@ <h3>Processing Model</h3>
 surface resource fetch events.</p>
 <div class="issue" data-number="27"></div>
 <ol data-link-for="PerformanceResourceTiming">
+<li>If the resource's <a data-cite="Fetch#concept-request">Request</a>'s
+<a data-cite="Fetch#concept-request-destination">destination</a> equals to
+"document", "object" or "embed", and was not triggered by
```
bzbarsky: If the destination is "object" or "embed", then the request definitely comes from the attribute-processing algorithms.   Those are the only things that set  destination to "object" or "embed".   So we only need to special-case the "document" case here, I think.

index.html
annevk: @bzbarsky my rough idea was that each global (client) has its own fetch group and since iframe always creates a global it'd have its own (embed and object are weird here and would require some special casing of sorts; my idea was that if they end up creating a global we'd move the fetches).

bzbarsky: That makes sense to me.  In that world, the load for  an `<iframe src="foo">` would go in the  fetch group of the iframe's `about:blank` global.  But for resource timing purposes we want to  add it to the resource timing of the parent page.

annevk: Yeah, I think the way resource timing does its tracking is somewhat unfortunate as we cannot hang it on this concept. We end up with another table in already complex (and undefined) landscape. Basically requires more special code paths to be carefully vetted each time we touch things.

bzbarsky: I think we should be able to hang resource timing on the **client** of the Request.  Would that work?

annevk: I don't think so. With `<iframe src=...>` the `about:blank` global is the client (and with which the service worker is associated).

bzbarsky: > With `<iframe src=...>` the about:blank global is the client

In this case we start with  https://html.spec.whatwg.org/multipage/iframe-embed-object.html#process-the-iframe-attributes then go to https://html.spec.whatwg.org/multipage/iframe-embed-object.html#otherwise-steps-for-iframe-or-frame-elements which does a https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigate and says:

> Any navigation required of the user agent in the process the iframe attributes algorithm must use the iframe element's node document's browsing context as the source browsing context.

When we land in the navigate algorithm, we end up in https://html.spec.whatwg.org/multipage/browsing-the-web.html#process-a-navigate-fetch which does:

> Set request's client to sourceBrowsingContext's active document's relevant settings object

How does the client end up as that `about:blank` global?  What am I missing?

annevk: Okay, I'm a little skeptical this is all agreed upon across implementations, but maybe this is all mostly correct and then it would probably work, indeed.

index.html
yoavweiss: So, maybe something like """
* Resource requests which client is not null and which destination is not "document", "object" or "embed" should be added to the request client's relevant global object's performance timeline.
* Resource requests which destination is "document",  which were triggered by the "process the iframe attribute" step, and which client has a document which has a browsing context which has a parent browsing context, should be added to the request's client's document's browsing context's parent browsing context's relevant global object's performance timeline
"""

There's probably a better way to phrase this, but let me know if that would properly define what we need.

bzbarsky: Things with destinations of "object" or "embed" should always be added, no?

I'm not sure I understand why we need all the conditions listed in the "document" case.  But going from request client to browsing context to parent browsing context looks wrong to me, per the discussion above: the client is already the settings object of the parent.

A settings object has no "relevant global".  I'm not sure what the best way in spec terms is to go from a settings object to the correct performance timeline object.  @domenic, do you know?

annevk: https://html.spec.whatwg.org/#concept-settings-object-global (if they're indeed associated with global objects).

yoavweiss: A performance timeline is (kinda) associated with a global object: https://w3c.github.io/performance-timeline/#performance-timeline

At the same time, it also seemed to be owned by an ES global environment. That's probably a mistake and it should be owned by https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-global

yoavweiss: OK, not sure how, but missed [the conclusion  from the other thread](https://github.com/w3c/resource-timing/pull/177#discussion_r230745051) when writing my previous comment here.

In that light, would a simple "resource Requests which client is not null should be added to the request client's Realm's global object's performance timeline" work?


> Things with destinations of "object" or "embed" should always be added, no?

Indeed. Got myself confused.

domenic: I don't fully understand what's going on here but I do want to say that you should never have to go through all three levels of the 1:1:1 concepts to get to each other. So

> request client's Realm's global object's

can just be

> request client's global object's

See https://html.spec.whatwg.org/multipage/webappapis.html#realms-settings-objects-global-objects whose second paragraph lists the definitions for navigating from each of realm/settings object/global object to each other. (In this case you'd link "global object" to https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-global .)

bzbarsky: > would a simple "resource Requests which client is not null should be added to the request client's Realm's global object's performance timeline" work

No, because for the "document" case you want to censor loads not triggered by https://html.spec.whatwg.org/multipage/obsolete.html#process-the-frame-attributes or https://html.spec.whatwg.org/multipage/iframe-embed-object.html#process-the-iframe-attributes afaict.

yoavweiss: I may very well be wrong, but isn't the "incumbent settings object" in [location object navigate](https://html.spec.whatwg.org/multipage/history.html#location-object-navigate) the settings object of the iframe's browsing context (rather than its parent)?
Similarly in step 13 of [shared-declarative-refresh-steps](https://html.spec.whatwg.org/#shared-declarative-refresh-steps), isn't the source browsing context the iframe's document browsing context?

bzbarsky: > isn't the "incumbent settings object" in location object navigate the settings object of the iframe's browsing context (rather than its parent)?

The incumbent settings object is the place where the `href` set or `assign` call happens.  It could be anything at all, including the iframe itself, its parent, or some other window  that has its hands on the Location object.

> Similarly in step 13 of shared-declarative-refresh-steps, isn't the source browsing context the iframe's document browsing context?

Yes it is.  I'm not sure why that matters here...   we want to exclude the loads done by https://html.spec.whatwg.org/#shared-declarative-refresh-steps from resource timing, right?

yoavweiss: > > isn't the "incumbent settings object" in location object navigate the settings object of the iframe's browsing context (rather than its parent)?
> 
> The incumbent settings object is the place where the `href` set or `assign` call happens. It could be anything at all, including the iframe itself, its parent, or some other window that has its hands on the Location object.

OK, thanks for correcting.

> > Similarly in step 13 of shared-declarative-refresh-steps, isn't the source browsing context the iframe's document browsing context?
> 
> Yes it is. I'm not sure why that matters here... we want to exclude the loads done by https://html.spec.whatwg.org/#shared-declarative-refresh-steps from resource timing, right?

We want to exclude it from the parent's resource timing, so for this case it won't be added to it right?

Regardless, I'll add the conditions to only include "document" destinations triggered by "process the iframe-attribute"

bzbarsky: > We want to exclude it from the parent's resource timing

Right.  That would be handled if we only add navigations that are caused by processing frame or iframe attributes, afaict...

yoavweiss: aside: should the same exclusions be added to prevent Service Workers from observing the same URLs?

bzbarsky: I haven't been following the service worker stuff closely, but my understanding is that generally navigations _do_ want to be seen by service workers, but how one determines which service worker should see the navigation is not entirely clear to me and is not really related to  resource timing.

For example, in the case of  `<iframe src="some-cross-origin-url">` we want the cross-origin url to show up in resource timing on the parent page of the iframe but the service worker that  gets used should be the one for  the target origin, not the one for the page the iframe is in, right?

yoavweiss:
@bzbarsky @annevk - Rewritten based on discussions. PTAL?

index.html
```diff
@@ -922,18 +926,25 @@ <h3>Processing Model</h3>
 <!-- Source: https://docs.google.com/document/d/1I7XGNJ57Qgjkg9pL11s7MK7zGEcwAgdNj1W5f7NKbW8/ -->
 <img src="timestamp-diagram.svg" alt="Resource Timing attributes"
 style='margin-top: 1em'></figure>
-<p>For each resource <a data-cite="FETCH#concept-fetch">fetched</a>
-by the current <a data-cite="HTML#browsing-context">browsing
-context</a>, excluding resources fetched by cross-origin
-stylesheets fetched with <code>no-cors</code> policy, perform the
-following steps:</p>
+
+<p>For each resource which <a>Request</a> has a non-null <a>client</a>, perform
```
bzbarsky: s/which/whose/?

index.html
```diff
@@ -922,18 +926,25 @@ <h3>Processing Model</h3>
...
 <p class="issue">Above cross-origin exclusion should be defined via
 Fetch registry: CSS needs to be defined in terms of Fetch and set
 some kind of "opaque request flag" for no-CORS CSS subresources. In
 turn, Resource Timing should interface with Fetch registry to
 surface resource fetch events.</p>
 <div class="issue" data-number="27"></div>
-<ol data-link-for="PerformanceResourceTiming">
+<li>If the resource's <a>Request</a>'s
+<a data-cite="Fetch#concept-request-destination">destination</a> equals to
+"document", and was not triggered by
```
bzbarsky: "and Request was not triggered", right?  The current wording is talking  about the destination being triggered...

index.html

index.html
yoavweiss: yup

yoavweiss:
Thanks for reviewing! :)

PR #183, opened 2018-12-12
Add a note regarding negative timestamps for navigation preload requests

### LINKED ISSUE #139: Navigation Preload can have negative times
As discussed in https://github.com/w3c/resource-timing/issues/88#issuecomment-297977711, Navigation Preload requests can (and likely to) start before the SW time origin. If they appear in the SW's timeline, they will have negative times.

We should either decide they do not appear in the SW's timeline and make that explicit, and somehow otherwise resolve this issue.

yoavweiss:
Closes #139

### DIFF
--- a/index.html
+++ b/index.html
@@ -484,6 +484,11 @@ required to service the request.</li>
 </li>
 <li>zero, otherwise.</li>
 </ol>
+<p class="note">Note that according to the definition of <a
+data-cite="HR-TIME-2#dfn-time-origin">time origin</a> in workers, it is possible
+and expected that certain attributes of <a
+href="https://w3c.github.io/ServiceWorker/#dom-serviceworkerregistration-navigationpreload">
+navigation preload</a> requests may include negative timestamps.</p>
 <p data-dfn-for="PerformanceResourceTiming">On getting, the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -484,6 +484,11 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
 </li>
 <li>zero, otherwise.</li>
 </ol>
+<p class="note">Note that according to the definition of <a
+data-cite="HR-TIME-2#dfn-time-origin">time origin</a> in workers, it is possible
+and expected that certain attributes of <a
+data-cite="service-workers-nightly#dom-serviceworkerregistration-navigationpreload">
+navigation preload</a> requests may include negative timestamps.</p>
```
cvazac: s/may include negative timestamps/will be negative `DOMHighResTimeStamp`s

yoavweiss: maybe "it is possible that some attributes of navigation preload requests will contain negative DOMHighResTimeStamp values"?

cvazac: or:
s/contain/have
"it is possible that some attributes of navigation preload requests will have negative DOMHighResTimeStamp values"

yoavweiss: sold!

PR #185, opened 2018-12-18
Connection reuse does not imply DNS was not required

### LINKED ISSUE #160: DNS lookups can occur even when reusing a connection
The spec currently says that domainLookupStart must return "The same value as fetchStart, if a persistent connection [RFC7230] is used or the resource is retrieved from relevant application caches or local resources." Similar for domainLookupEnd. However, there could be a DNS lookup even when reusing a connection, so the spec currently allows (domainLookupEnd - domainLookupStart) to be 0 in some cases where a DNS lookup actually occurs. This feels wrong.

More context: see Chromium [bug](https://crbug.com/760059).

yoavweiss:
Closes #160 
As discussed on the call, this is not currently testable in WPT infrastructure. I'm planning to open an issue to keep track of tests for this.

***
<a href="https://pr-preview.s3.amazonaws.com/yoavweiss/resource-timing/pull/185.html" title="Last updated on Dec 19, 2018, 11:24 PM UTC (78bd09f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/185/13d5a2e...yoavweiss:78bd09f.html" title="Last updated on Dec 19, 2018, 11:24 PM UTC (78bd09f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -524,11 +524,12 @@ otherwise.</li>
 <p data-dfn-for="PerformanceResourceTiming">On getting, the
 <dfn>domainLookupStart</dfn> attribute MUST return as follows:</p>
 <ol data-link-for="PerformanceResourceTiming">
-<li>The same value as <a>fetchStart</a>, if a <a href=
+<li>The same value as <a>fetchStart</a>, if no domain lookup was
+required to fetch the resources (e.g. due to the fact that a <a href=
 "https://tools.ietf.org/html/RFC7230#section-6.3">persistent
-connection</a> [[RFC7230]] is used or the resource is retrieved
-from <a data-cite="HTML#relevant-application-cache">relevant
-application caches</a> or local resources.</li>
+connection</a> [[RFC7230]] was used or in case the resource was
+retrieved from <a data-cite="HTML#relevant-application-cache">relevant
+application caches</a> or local resources).</li>
 <li>The time immediately after the user agent before the domain
 data retrieval from the domain information cache, if the user agent
 has the domain information in cache.</li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -524,11 +524,12 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
 <p data-dfn-for="PerformanceResourceTiming">On getting, the
 <dfn>domainLookupStart</dfn> attribute MUST return as follows:</p>
 <ol data-link-for="PerformanceResourceTiming">
-<li>The same value as <a>fetchStart</a>, if a <a href=
+<li>The same value as <a>fetchStart</a>, if no domain lookup was
+required to fetch the resources (e.g. due to the fact that a <a href=
```
cvazac: s/due to the fact that/if

index.html
```diff
@@ -524,11 +524,12 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
...
 "https://tools.ietf.org/html/RFC7230#section-6.3">persistent
-connection</a> [[RFC7230]] is used or the resource is retrieved
-from <a data-cite="HTML#relevant-application-cache">relevant
-application caches</a> or local resources.</li>
+connection</a> [[RFC7230]] was used or in case the resource was
+retrieved from <a data-cite="HTML#relevant-application-cache">relevant
+application caches</a> or local resources).</li>
```
cvazac: do we also need this change the other places we mention "persistent connection"?

yoavweiss: Not all of them, but I did leave out domainLookupEnd...

Issue #173, opened 2018-10-15
Clarify cross-origin resources


alanwaketan:
When the spec refers cross-origin resources, are we talking about response or request origin?
For example, in resource_timing_cross_origin_redirect_chain.html, the resource loaded in frameContext was requested from a cross-origin site but then the respond is from the same-origin site. Should we treat the resource as cross-origin or same-origin? The test itself suggests the former. However, I didn't see any spec language defining this.

alanwaketan:
Thanks Anne for clarification. Let's say if multiple redirects happen, what origins are required in TAO? Also do we need TAO for each intermediate response?

Let's say:
A.com -> B.com -> C.com
1. C.com has TAO B.com, B.com has TAO A.com, is this the only case that goes through?
2. Or as long as C.com has TAO A.com, it will go through?

alanwaketan:
May I ask why resource_TAO_cross_origin_redirect_chain.html would pass? The last redirect doesn't have TAO. Can you clarify where the TAO needs to be present in the chain?

annevk:
If multiple redirects happen, the origin gets tainted and will be send out as "null". C should see "null" and reply accordingly. It shouldn't see B.com. And yes, each response in the chain needs to have TAO, unless there are some same-origin responses and it hasn't gone cross-origin yet.

(The tests may well be incorrect.)

alanwaketan:
Thanks Anne for response. I am so sorry that it takes me a while to response.

I would like to write a few examples and see if I understand you correctly. All examples assume the main frame is on A.com, and the redirect chain in each example assumes an iframe within the main frame.

Example 1:
A.com => B.com (TAO) => A.com (TAO) => A.com (TAO)
The main frame should see the iframe's timing.

Example 2:
A.com => A.com => B.com (TAO) => A.com (TAO)
The main frame should see the iframe's timing.

Example 3:
A.com => B.com (TAO) => A.com (TAO) => A.com
The main frame should not see the iframe's timing.

Example 4:
A.com => B.com => A.com (TAO)
The main frame should not see the iframe's timing.

Example 5:
A.com => B.com => A.com
The main frame should not see the iframe's timing.

alanwaketan:
Thanks Anne for the clarification and sorry for late response.

I would like to give examples to see if I fully understand everything. Let's say we have a main frame in A.com, and some redirections happen within its child frame. How could the main frame get accesses to the child frame's timing information in each example?

Example 1.
[...]

PR #194, opened 2019-01-04
Prevent leaking values when TAO fails

yoavweiss:
This is an issue discovered due to a bug [introduced](https://chromium-review.googlesource.com/c/chromium/src/+/1363906/3/third_party/blink/renderer/core/timing/performance_resource_timing.cc) with secureConnectionStart handling.

While going over the spec, I realized that the "on getting" processing model does not always take TAO into consideration, and may leak some of the values. AFAICT this is mostly tested against, so it's just a spec issue when it comes to TAO. https://github.com/web-platform-tests/wpt/pull/14719 tests that secureConnectionStart values are not fetchStart when the connection is not secure (which exposed the Chrome issue).

### DIFF
--- a/index.html
+++ b/index.html
@@ -531,24 +531,27 @@ otherwise.</li>
 <p data-dfn-for="PerformanceResourceTiming">On getting, the
 <dfn>domainLookupStart</dfn> attribute MUST return as follows:</p>
 <ol data-link-for="PerformanceResourceTiming">
+<li>Zero, if the last non-redirected
+<a data-cite="FETCH#concept-fetch">fetch</a> of the resource does not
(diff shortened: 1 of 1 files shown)


npm1:
Ahh thanks for this fix, makes much more sense! I do have a question, now that we're looking at this: is checking TAO for the last redirect sufficient? Wouldn't we need to check TAO for the whole redirect chain?

yoavweiss:
That's a very good point and will address (or handwave) some of the concerns raised in https://github.com/w3c/resource-timing/issues/173 and https://github.com/w3c/resource-timing/issues/152.

yoavweiss:
Opened https://github.com/w3c/resource-timing/pull/196 to address this.

PR #209, opened 2019-05-14
Split response end to a separate algorithm

### LINKED ISSUE #208: Encapsulate responseEnd in an algorithm
The [ElementTiming](https://github.com/WICG/element-timing) API would like to expose responseEnd as well, which means it would be nice if ResourceTiming encapsulates the responseEnd description into an algorithm. This way, ElementTiming can just hook into it by providing the appropriate arguments into the algorithm.

yoavweiss:
Closes #208

### DIFF
--- a/index.html
+++ b/index.html
@@ -674,15 +674,20 @@ a request, due to an <a data-cite=
 the last request.</p>
 </aside>
 <p data-dfn-for="PerformanceResourceTiming">On getting, the
-<dfn>responseEnd</dfn> attribute MUST return as follows:</p>
+<dfn>responseEnd</dfn> attribute MUST return the result of running
+<a>get response end time</a> algorithm.</p>
+
+<p>When requested to run the <dfn>get response end time</dfn>
+algorithm, the user agent must run the following steps:</p>
 <ol>
-<li>The time immediately after the user agent receives the last
-byte of the response or immediately before the transport connection
-is closed, whichever comes first. The resource here can be received
-either from <a data-cite="HTML#relevant-application-cache">relevant
-application caches</a>, local resources, or from the server.</li>
-<li>The time immediately before the user agent aborts the fetch due
-to a network error.</li>
+<li>If the fetch was aborted due to a network error, return the time
+immediately before the user agent aborts the fetch.</li> 
+<li>Otherwise, return the time immediately after the user agent
+receives the last byte of the response or immediately before the
+transport connection is closed, whichever comes first. The resource
+here can be received either from <a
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -674,15 +674,20 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
 the last request.</p>
 </aside>
 <p data-dfn-for="PerformanceResourceTiming">On getting, the
-<dfn>responseEnd</dfn> attribute MUST return as follows:</p>
+<dfn>responseEnd</dfn> attribute MUST return the result of running
```
npm1: nit: of running the algorithm to <a>get ...

index.html
```diff
@@ -674,15 +674,20 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
...
-byte of the response or immediately before the transport connection
-is closed, whichever comes first. The resource here can be received
-either from <a data-cite="HTML#relevant-application-cache">relevant
-application caches</a>, local resources, or from the server.</li>
-<li>The time immediately before the user agent aborts the fetch due
-to a network error.</li>
+<li>If the fetch was aborted due to a network error, return the time
+immediately before the user agent aborts the fetch.</li> 
+<li>Otherwise, return the time immediately after the user agent
+receives the last byte of the response or immediately before the
```
npm1: This can be more algorithmic:
Let timeA be ...
Let timeB be...
Return the mininum between timeA and timeB.

index.html
```diff
@@ -674,15 +674,20 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
...
+transport connection is closed, whichever comes first. The resource
```
npm1: Is this last sentence needed, and if so, is it normative text or just a clarification note?

index.html
yoavweiss: I agree that it can, but not sure it's of huge value as the times in question are not well-defined, and we will need to better define them through Fetch integration.

index.html
yoavweiss: Again, this is something that will be clarified through Fetch integration. For now I preferred to keep most of the original text.

index.html
yoavweiss: sure

index.html
npm1: It has the value of improving the spec 'health' a little bit and more clarity. But I want this change to land anyways, so it's fine if we keep as is for now.

yoavweiss: I agree, but I expect this to change significantly (and don't think rewriting it now would add significant short-term value).

PR #217, opened 2019-11-20
Protect workerStart with a TAO check

### LINKED ISSUE #118: should redirected navigations have a workerStart value if a previous redirect URL was intercepted
Consider the following situation:

1. Use opens tab a.com
2. a.com is intercepted by a service worker which results in a redirect to b.com
3. b.com is loaded

Two questions:

a. Should there be a `.workerStart` for the a.com interception?  There would if this was a subresource request.
b. Should the `.workerStart` value be controlled by the `Timing-Allow-Origin` header?  The redirectStart/redirectEnd values are controlled by that header, but workerStart doesn't mention it.

If the answer to (a) is "yes", then consider:

1. a.com is intercepted by a SW and redirects to b.com
2. b.com is intercepted by a SW and redirects to c.com
3. c.com is loaded

Should the workerStart for a.com or b.com be shown?

yoavweiss:
Closes #118

***
<a href="https://pr-preview.s3.amazonaws.com/yoavweiss/resource-timing/pull/217.html" title="Last updated on Nov 28, 2019, 3:47 PM UTC (7d5ad55)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/217/70baa57...yoavweiss:7d5ad55.html" title="Last updated on Nov 28, 2019, 3:47 PM UTC (7d5ad55)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1069,7 +1069,8 @@ agent MUST run the following substeps:
   <ol>
     <li>If the resource's <a>Request</a>'s
     <a data-cite="Fetch#concept-request-destination">destination</a> equals to
-    "document", set <a>workerStart</a> to zero.</li>
+    "document", set <a>workerStart</a> to zero. Otherwise, abort these
+    substeps.</li>
     <li>Set <a>redirectStart</a>, <a>redirectEnd</a>,
     <a>domainLookupStart</a>, <a>domainLookupEnd</a>, <a>connectStart</a>,
     <a>connectEnd</a>, <a>requestStart</a>, <a>responseStart</a> and
--- a/index.html
+++ b/index.html
@@ -477,7 +477,8 @@ uses the ALPN Protocol ID's to indicate the protocol in use.</p>
 <li>
 <p>If the current browsing or worker context's have an
 <a data-cite="service-workers-1#dfn-containing-service-worker-registration">
-active worker</a> [[service-workers-1]]:</p>
+active worker</a> [[service-workers-1]] and the resource
+passes the <a>timing allow check</a> algorithm:</p>
 <ol>
 <li>the time immediately before the user agent <a data-cite=
 "service-workers-1#on-fetch-request-algorithm">fires an event named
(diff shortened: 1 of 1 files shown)


yoavweiss:
@npm1 - revamped to exclude non-document destinations. PTAL?

index.html
```diff
@@ -475,9 +475,12 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
 <dfn>workerStart</dfn> attribute MUST return as follows:</p>
 <ol>
 <li>
-<p>If the current browsing or worker context's have an
+<p>If the current browsing or worker context have an
```
npm1: Is there a better way to refer to this 'browsing or worker context'?

index.html
```diff
@@ -1050,8 +1053,23 @@ <h3>Processing Model</h3>
...
+be the same value as <a>fetchStart</a>.</li>
+<li><dfn data-lt="step-collection-start">If the user agent is to
+reuse the data from another existing or completed <a data-cite=
+"FETCH#concept-fetch">fetch</a> initiated from the <a>current
+document</a></dfn>, abort the remaining steps.</li>
+<li>If the resource's <a>Request</a>'s
+<a data-cite="Fetch#concept-request-destination">destination</a> equals to
+"document" and the resource fails the <a>timing allow check</a> algorithm, the
+user agent MUST set <a>workerStart</a> to zero and go to the step labeled
+<a href="#dfn-step-response-end">response end</a>.</li>
```
npm1: Why does this jump to response end step? In this case when would redirectStart and friends be set?

yoavweiss: You're right, changed the logic to merge the conditionals that lead to jumping to the response end.

index.html
yoavweiss: We need to rewrite this, probably as the 
[active service worker](https://html.spec.whatwg.org/#concept-environment-active-service-worker) of the [context object](https://dom.spec.whatwg.org/#context-object)'s
[relevant settings object](https://html.spec.whatwg.org/#relevant-settings-object).

I'll do that in a followup.

index.html
```diff
@@ -475,9 +475,12 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
 <dfn>workerStart</dfn> attribute MUST return as follows:</p>
 <ol>
 <li>
-<p>If the current browsing or worker context's have an
+<p>If the current browsing or worker context have an
 <a data-cite="service-workers-1#dfn-containing-service-worker-registration">
-active worker</a> [[service-workers-1]]:</p>
+active worker</a> [[service-workers-1]] and if the resource
```
npm1: nit: perhaps have the check on the destination type first, as that check is cheaper and that is probably how it would be implemented.

index.html
```diff
@@ -1050,8 +1053,31 @@ <h3>Processing Model</h3>
...
+<a>domainLookupEnd</a>, <a>connectStart</a> and <a>connectEnd</a>
+be the same value as <a>fetchStart</a>.</li>
+<li><dfn data-lt="step-collection-start">If the user agent is to
+reuse the data from another existing or completed <a data-cite=
+"FETCH#concept-fetch">fetch</a> initiated from the <a>current
+document</a></dfn>, abort the remaining steps.</li>
+<li>If the resource fails the <a>timing allow check</a> algorithm, the user
+agent MUST run the following substeps:
+  <ol>
+    <li>If the resource's <a>Request</a>'s
```
npm1: This logic is still leaving workerStart unset when the TAO check fails and destination is not "document". Perhaps we should move the service worker step so that it occurs before this step? This way, when TAO fails, we just overwrite the workerStart value to 0 when needed (and keep the computed value otherwise). Alternatively, create a var representing the TAO check. Set workerStart using this variable and the destination logic. Then, have the step that sets 0's and skips when this check fails.

yoavweiss: Forgot to bail if the destination is not document...

index.html
yoavweiss: Note that it's an "or" relationship, so both tests will need to run (unless I misunderstand and you want destination and TAO checks to run before the active SW one).

index.html
yoavweiss: Cleaned it up here. PTAL

Issue #39, opened 2015-10-02
Define Resource Timing in terms of Fetch


igrigorik:
Expressing RT in terms of Fetch will force us to clarify a lot of the current edge cases:
- reporting of redirects and failed fetches (see https://github.com/w3c/resource-timing/issues/12)
- reporting of multi-request fetches (see https://github.com/w3c/resource-timing/issues/21)
- initiator reporting (see https://github.com/w3c/resource-timing/pull/36)

Some ideas on mechanics of how to accomplish this in https://github.com/w3c/resource-timing/pull/36#issuecomment-143150479. This will be a substantial effort with required changes both in RT and Fetch specs.

plehegar:
See https://github.com/w3c/web-platform-tests/pull/2884/commits/e21f1ce62e87216960376b86c9d49ba282dd129d for test.

noamr:
I believe this can be closed.

yoavweiss:
Amazing work!!

PR #225, opened 2020-03-13
Pass realm when creating PerformanceResourceTiming

npm1:
Relevant issue: https://github.com/w3c/performance-timeline/issues/162.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/pull/225.html" title="Last updated on Mar 24, 2020, 8:20 PM UTC (3580f11)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/225/00f42ad...3580f11.html" title="Last updated on Mar 24, 2020, 8:20 PM UTC (3580f11)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -617,7 +617,7 @@ the connection to the server to retrieve the resource, otherwise.
   <li>If the user agent waits for full handshake completion to send the
     request, this interval includes the full TLS handshake even if other
     requests were sent using early data on this connection.</li>
-</ul> 
+</ul>
 
 <p class=note>Example: Suppose the user agent establishes an HTTP/2 connection
 over TLS 1.3 to send a GET request and a POST request. It sends the ClientHello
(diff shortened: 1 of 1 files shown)


npm1:
Ping

rniwa:
Hm... I'm not certain that grabbing the realm out of request's client is always right. We need someone familiar with fetch like @annevk to review this.

annevk:
I don't understand how "For each resource whose Request has a non-null client" fits into the web platform.

yoavweiss:
@annevk We're (painfully) aware that we are hand-wavily using "resource" here as if it's a thing, but this issue is somewhat tangential to that. If we assume that Request is properly acquired and has a non-null client, can we use its client's realm to create the performance entry?

annevk:
I'm not sure. Request's client concept has a number of unresolved issues so it really depends.

yoavweiss:
> I'm not sure. Request's client concept has a number of unresolved issues so it really depends.

@annevk - Do you see a path forward here before full integration of this spec with Fetch?

annevk:
#39 is from 2015. What is the ETA?

dudss37:
[]()

noamr:
I think this should be closed.

yoavweiss:
Agree

PR #236, opened 2020-09-29
Conform 'current time' definition to the current hr-time spec

igneel64:
Conform 'current time' definition to the current hr-time spec [current high resolution time](https://w3c.github.io/hr-time/#dfn-current-high-resolution-time). 

Added `HR-TIME-ED` localBiblio entry.

***
<a href="https://pr-preview.s3.amazonaws.com/igneel64/resource-timing/pull/236.html" title="Last updated on Sep 29, 2020, 2:16 PM UTC (f581ed7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/236/05e9c1b...igneel64:f581ed7.html" title="Last updated on Sep 29, 2020, 2:16 PM UTC (f581ed7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -63,6 +63,13 @@
     wgPatentURI:  "https://www.w3.org/2004/01/pp-impl/45211/status",
     wgPatentPolicy: "PP2017",
     xref: ["html", "hr-time-2", "performance-timeline-2"],
+    localBiblio: {
+      "HR-TIME-ED": {
+        title: "High Resolution Time Editor's Draft",
+        status: "ED",
+        href: "https://w3c.github.io/hr-time/", 
+      },
+    }
   };
 </script>
 </head>
@@ -216,9 +223,9 @@ milliseconds since the start of navigation of the document
 [[HR-TIME-2]]. For example, the <a data-cite=
 "NAVIGATION-TIMING-2#dom-PerformanceNavigationTiming-startTime">start
 of navigation of the document</a> occurs at time 0.</p>
-<p>The term <dfn>current time</dfn> refers to the number of
-milliseconds since the start of navigation of the document until
-the current moment in time.</p>
+<p>The term <dfn>current time</dfn> refers to the result of running 
+<a data-cite="HR-TIME-ED#dfn-current-high-resolution-time">current high 
+resolution time</a> where |current global| is the {{Window}} object.</p>
 <p class='note'>This definition of time is based on the High
 Resolution Time specification [[HR-TIME-2]] and is different from
 the definition of time used in the Navigation Timing specification


index.html
```diff
@@ -216,9 +223,9 @@ <h2>Terminology</h2>
 [[HR-TIME-2]]. For example, the <a data-cite=
 "NAVIGATION-TIMING-2#dom-PerformanceNavigationTiming-startTime">start
 of navigation of the document</a> occurs at time 0.</p>
-<p>The term <dfn>current time</dfn> refers to the number of
-milliseconds since the start of navigation of the document until
-the current moment in time.</p>
+<p>The term <dfn>current time</dfn> refers to the result of running 
+<a data-cite="HR-TIME-ED#dfn-current-high-resolution-time">current high 
+resolution time</a> where |current global| is the {{Window}} object.</p>
```
yoavweiss, line 228: Maybe you can add an inline issue (`<p class=issue>` IIRC) to align this to the published spec, once we've actually published it

index.html
```diff
@@ -216,9 +223,9 @@ <h2>Terminology</h2>
 [[HR-TIME-2]]. For example, the <a data-cite=
 "NAVIGATION-TIMING-2#dom-PerformanceNavigationTiming-startTime">start
 of navigation of the document</a> occurs at time 0.</p>
-<p>The term <dfn>current time</dfn> refers to the number of
-milliseconds since the start of navigation of the document until
-the current moment in time.</p>
+<p>The term <dfn>current time</dfn> refers to the result of running
```
yoavweiss, line 226: I think we'd be better off eliminating this dfn, and call the algorithm directly with the current global object from the few places in the processing model that are calling it. 

The reason for that is that we need to call it with a specific global object, and this dfn doesn't give us that. Does that make sense?

igneel64: Sounds good and also seems more robust since I have observed that the term `current time` is mostly used in algorithmic steps in this spec.

So would replacing occurences like:
`record the <a>current time</a> in <a>startTime</a>`
with
`record the result of <a data-cite="HR-TIME-ED#dfn-current-high-resolution-time">current high 
resolution time</a> where |current global| is {{Window}} in <a>startTime</a>`
fit the case ?

index.html
igneel64: Probably not needed as this will be on L3, right ?

index.html
yoavweiss: `{{Window}}` is a reference to the Window object, but not to a specific window.
You probably want to reference the [current global object](https://html.spec.whatwg.org/multipage/webappapis.html#current-global-object)

index.html
yoavweiss: indeed!

yoavweiss:
@igneel64 - Friendly ping! I'd love to help you land this..

mmocny:
Issue triage: this needs a rebase, and someone to validate that the issue remains and still needs a fix.

Sounds like a good task for an LLM!

Issue #244, opened 2021-01-18
Definition of `name` needs to be more precise


yoavweiss:
"This attribute MUST return the resolved URL of the requested resource. This attribute MUST NOT change even if the fetch redirected to a different URL." is hand-wavy. We need to define it in precise Fetch terms that make it clear that redirected URLs are not exposed (as well as clarify it as a note)

noamr:
I believe this can be closed.

Issue #252, opened 2021-01-18
4.6 Integrate the processing model with Fetch


yoavweiss:
* [x] The algorithm should be called (probably from Fetch) with clear inputs.
* [x] Step 1 - need to replace "fetched by a cross-origin stylesheet fetch with no-cors policy" with defined concepts related to the Request object.
* [x] Step 2 - replace "triggered by..." with well-defined concepts related to the Request object
* [x] Step 4 - startTime needs to be recorded as part of Fetch and communicated to the processing mode
* [x] Step 5 - The processing model of initiatorType needs to be defined in terms of Fetch. See https://github.com/w3c/resource-timing/issues/246
* [x] Step 6 - define in terms of Fetch. See https://github.com/w3c/resource-timing/issues/244
* [x] Step 7 - fetchStart needs to be recorded as part of Fetch and communicated to the processing mode
* [x] Step 8 - Seems spurious if a fetch is not initiated or terminated. Needs to fall out from the integration between Fetch and this processing model. May require definition of the Memory/Preload Cache. See https://github.com/whatwg/fetch/issues/590
* [x] Step 9 - workerStart definition should be set by Fetch/ServiceWorkers and communicated to this processing model
* [x] Step 12 - replace "resource is fetched from" with well-defined concepts
* [x] Steps 13-15 - domain lookup and connection times are not currently defined in Fetch. Need to investigate to see if/how they can be better defined.
* [x] Step 16 - "secure transport is used" needs to be replaced with Fetch concepts. Better definition of the connection times needs to be investigated.
* [x] Step 17-19 - Set requestStart, responseStart and responseEnd of part of Fetch processing.
* [x] Step 19.2 - define transferSize, encodedBodySize and decodedBodySize in terms of Fetch. Note that on transferSize, we may want to omit the header size: https://github.com/w3c/resource-timing/issues/238
* [x] Step 21 - Define redirect processing as part of the Fetch processing model
* [x] Define if entry is queued before or after the load event https://github.com/w3c/resource-timing/issues/82

noamr:
I'm seeing to ways to possible ways to go about this:

1. keep the data as a struct inside `fetch`, and call `resource timing` only at the end of the response with all the info. This would basically mean keeping a lot of timing information inside the fetch spec.
2. Call `resource timing` from multiple locations with different parameters.

I'm currently leaning towards (2), but would love to run it by you first, @annevk and @yoavweiss

annevk:
[...]

PR #257, opened 2021-02-28
Use FETCH#URL to describe the name of a request

noamr:
Instead of specifying here that it's the URL before redirects,
which is essentially the same thing.

Closes https://github.com/w3c/resource-timing/issues/244

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/pull/257.html" title="Last updated on Feb 28, 2021, 11:28 AM UTC (f3f4bf0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/257/05e9c1b...f3f4bf0.html" title="Last updated on Feb 28, 2021, 11:28 AM UTC (f3f4bf0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -365,9 +365,9 @@ interface:</p>
 <dl data-link-for="PerformanceResourceTiming">
 <dt><dfn>name</dfn></dt>
 <dd>This attribute MUST return the <a data-cite=
-"HTML#resolve-a-url">resolved URL</a> of the requested resource.
-This attribute MUST NOT change even if the <a data-cite=
-"FETCH#concept-fetch">fetch</a> redirected to a different URL.</dd>
+"FETCH#concept-request-url">URL</a> of the resource's <a data-cite=
+"FETCH#concept-request">request</a>.
+<p class="note">This is the URL requested originally, and is not affected by redirects.</dd>
 <dt><dfn>entryType</dfn></dt>
 <dd>The <a>entryType</a> attribute MUST return the DOMString
 "<code>resource</code>".</dd>


index.html
```diff
@@ -365,9 +365,9 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
 <dl data-link-for="PerformanceResourceTiming">
 <dt><dfn>name</dfn></dt>
 <dd>This attribute MUST return the <a data-cite=
-"HTML#resolve-a-url">resolved URL</a> of the requested resource.
-This attribute MUST NOT change even if the <a data-cite=
-"FETCH#concept-fetch">fetch</a> redirected to a different URL.</dd>
+"FETCH#concept-request-url">URL</a> of the resource's <a data-cite=
```
yoavweiss, line 368: Looking at https://fetch.spec.whatwg.org/#concept-request-url I don't see it being explicitly set to the first URL in the list, and there's a not saying "implementations are encouraged" to do that. As such, I don't know if this is something we want to rely on. Maybe we should explicitly point to the first URL in the [URLList](https://fetch.spec.whatwg.org/#concept-request-url-list)?

/cc @annevk for his opinions

noamr: Seems like `URL` is set to be exactly that, judging by the fact that during redirects it's not changed.
But I would let @annevk confirm.

yoavweiss: Yeah, looking closer it seems to be set in the Request constructor and not modified on redirects.

annevk:
I'm not sure what "resource" is, but I suspect it's a response. Assuming I'm right, I think you want to get data from the response and not assume that the response has access to the request.

yoavweiss:
> I'm not sure what "resource" is

There's a plan to clarify that as part of https://github.com/w3c/resource-timing/issues/252, but I don't know we have a full design for that just yet.

> I suspect it's a response. Assuming I'm right, I think you want to get data from the response and not assume that the response has access to the request.

Would it make sense then to grab the first item from  [https://fetch.spec.whatwg.org/#concept-response-url-list](Response's URL list)?

annevk:
Yeah, probably, if you can be certain it's non-empty (i.e., you do not access it when fetch hasn't returned it yet).

noamr:
> > I'm not sure what "resource" is
> 
> There's a plan to clarify that as part of #252, but I don't know we have a full design for that just yet.

It might be good to start from #252 then, it would clarify a lot of these issues, as the request/response etc will probably be a parameter passed from FETCH at a particular point in time.

PR #261, opened 2021-03-17
Prepare for FETCH integration

noamr:
- Remove duplicates that are moved to FETCH
- Add an exported algorithm to mark resource timing

See https://github.com/whatwg/fetch/pull/1185 and https://github.com/w3c/resource-timing/issues/252

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/pull/261.html" title="Last updated on Apr 2, 2021, 6:15 AM UTC (ce74048)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/261/bd74994...ce74048.html" title="Last updated on Apr 2, 2021, 6:15 AM UTC (ce74048)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -646,9 +646,41 @@ that <code>resourcetimingbufferfull</code> event handlers call
 </li>
 </ol>
 </section>
-<section id="sec-cross-origin-resources">
+<section id="sec-timing-allow-origin"> 
+ <h4><code>Timing-Allow-Origin</code> Response Header</h4>
+<p class="note">This section is non-normative.</p>
+ <p>The <dfn>Timing-Allow-Origin</dfn> HTTP response header field
+ can be used to communicate a policy indicating origin(s) that are
+ allowed to see values of attributes that would have been zero due
+ to the cross-origin restrictions. The header's value is represented
+ by the following ABNF [[RFC5234]] (using <a data-cite=
+ "RFC7230#section-7">List Extension</a>, [[RFC7230]]):</p>
+ <pre class="abnf">
+       Timing-Allow-Origin = 1#( <a data-cite=
+ "FETCH#origin-header">origin-or-null</a> / <a data-cite=
+ "FETCH#http-new-header-syntax">wildcard</a> )
+     </pre>
+ <p>The sender MAY generate multiple <a>Timing-Allow-Origin</a>
+ header fields. The recipient MAY combine multiple
+ <a>Timing-Allow-Origin</a> header fields by appending each
+ subsequent field value to the combined field value in order,
+ separated by a comma.</p>
+ <p>The <dfn>timing allow check</dfn> algorithm, which checks
+whether a resource's timing information can be shared with the
+ <a>current document</a>, is as follows:</p>
+ <ol>
+     <li>Let <var>response</var> be the resource's <a data-cite=
+       "FETCH#concept-response">Response</a>.</li>
+     <li>Return <var>response</var>'s <a data-cite=
+       "FETCH#concept-response-timing-allow-passed">
+       timing allow passed flag</a>.</li>
+ </ol>
+ <p class=note>The Timing-Allow-Origin header may arrive as part of a cached
+ response. In case of cache revalidation, according to
+ <a href="https://tools.ietf.org/html/rfc7234#section-4.3.4">RFC 7234</a>,
+ the header's value may come from the revalidation response, or if not present
+ there, from the original cached resource.</p>
  <h3>Cross-origin Resources</h3>
- <p>This section is non-normative.</p>
  <p data-dfn-for="PerformanceResourceTiming">As detailed in [=fetch=],
  cross-origin resources are included as <a>PerformanceResourceTiming</a> objects in the
  <a data-cite="PERFORMANCE-TIMELINE-2#performance-timeline">Performance
@@ -669,7 +701,6 @@ that <code>resourcetimingbufferfull</code> event handlers call
  <a>cross-origin</a> restrictions previously specified in this
  section.</p>
 </section>
- 
 <section id="attribute-descriptions">
  <h3>Resource Timing Attributes</h3>
  <p>This section is non-normative.</p>
--- a/index.html
+++ b/index.html
@@ -646,6 +646,28 @@ that <code>resourcetimingbufferfull</code> event handlers call
 </li>
 </ol>
 </section>
+
+<section id="sec-cross-origin-resources">
+ <h3>Cross-origin Resources</h3>
+ <p data-dfn-for="PerformanceResourceTiming">As detailed in [=fetch=],
+ cross-origin resources are included as <a>PerformanceResourceTiming</a> objects in the
+ <a data-cite="PERFORMANCE-TIMELINE-2#performance-timeline">Performance
+ Timeline</a>. If the <a href="FETCH#concept-tao-check">timing allow check</a> algorithm
+ fails for a resource, these attributes of its <a>PerformanceResourceTiming</a>
+ object are set to zero:
+{{PerformanceResourceTiming/redirectStart}}, {{PerformanceResourceTiming/redirectEnd}},
+{{PerformanceResourceTiming/domainLookupStart}}, {{PerformanceResourceTiming/domainLookupEnd}},
+{{PerformanceResourceTiming/connectStart}}, {{PerformanceResourceTiming/connectEnd}},
+{{PerformanceResourceTiming/requestStart}}, {{PerformanceResourceTiming/responseStart}},
+{{PerformanceResourceTiming/secureConnectionStart}}, {{PerformanceResourceTiming/transferSize}},
+{{PerformanceResourceTiming/encodedBodySize}}, and {{PerformanceResourceTiming/decodedBodySize}}.</p>
+ <p>Server-side applications may return the
+ <a data-cite="FETCH#concept-tao-check">Timing-Allow-Origin</a>
+ HTTP response header to allow the User
+ Agent to fully expose, to the document origin(s) specified, the
+ values of attributes that would have been zero due to the
+ <a>cross-origin</a> restrictions previously specified in this
+ section.</p>
 <section id="sec-timing-allow-origin"> 
  <h4><code>Timing-Allow-Origin</code> Response Header</h4>
 <p class="note">This section is non-normative.</p>
@@ -680,26 +702,6 @@ whether a resource's timing information can be shared with the
  <a href="https://tools.ietf.org/html/rfc7234#section-4.3.4">RFC 7234</a>,
  the header's value may come from the revalidation response, or if not present
  there, from the original cached resource.</p>
- <h3>Cross-origin Resources</h3>
- <p data-dfn-for="PerformanceResourceTiming">As detailed in [=fetch=],
- cross-origin resources are included as <a>PerformanceResourceTiming</a> objects in the
- <a data-cite="PERFORMANCE-TIMELINE-2#performance-timeline">Performance
- Timeline</a>. If the <a href="FETCH#concept-tao-check">timing allow check</a> algorithm
- fails for a resource, these attributes of its <a>PerformanceResourceTiming</a>
- object are set to zero:
-{{PerformanceResourceTiming/redirectStart}}, {{PerformanceResourceTiming/redirectEnd}},
-{{PerformanceResourceTiming/domainLookupStart}}, {{PerformanceResourceTiming/domainLookupEnd}},
-{{PerformanceResourceTiming/connectStart}}, {{PerformanceResourceTiming/connectEnd}},
-{{PerformanceResourceTiming/requestStart}}, {{PerformanceResourceTiming/responseStart}},
-{{PerformanceResourceTiming/secureConnectionStart}}, {{PerformanceResourceTiming/transferSize}},
-{{PerformanceResourceTiming/encodedBodySize}}, and {{PerformanceResourceTiming/decodedBodySize}}.</p>
- <p>Server-side applications may return the
- <a data-cite="FETCH#concept-tao-check">Timing-Allow-Origin</a>
- HTTP response header to allow the User
- Agent to fully expose, to the document origin(s) specified, the
- values of attributes that would have been zero due to the
- <a>cross-origin</a> restrictions previously specified in this
- section.</p>
 </section>
 <section id="attribute-descriptions">
  <h3>Resource Timing Attributes</h3>
--- a/index.html
+++ b/index.html
@@ -687,15 +687,15 @@ that <code>resourcetimingbufferfull</code> event handlers call
  there, from the original cached resource.</p>
 </section>
 <section id="sec-iana-considerations">
- <h4>IANA Considerations</h4>
- <p>This section registers <a>Timing-Allow-Origin</a> as a <a href="https://tools.ietf.org/html/rfc3864#section-4.2.2">Provisional Message Header</a>.</p>
- <dl>
- <dt>Header field name:</dt><dd><pre class="abnf">Timing-Allow-Origin</pre></dd>
- <dt>Applicable protocol:</dt><dd>http</dd>
- <dt>Status:</dt><dd>provisional</dd>
- <dt>Author/Change controller:</dt><dd><a href="https://www.w3.org/">W3C</a></dd>
- <dt>Specification document:</dt><dd><a href="#sec-timing-allow-origin"></a></dd>
- </dl>
+<h4>IANA Considerations</h4>
+<p>This section registers <a>Timing-Allow-Origin</a> as a <a href="https://tools.ietf.org/html/rfc3864#section-4.2.2">Provisional Message Header</a>.</p>
+<dl>
+<dt>Header field name:</dt><dd><pre class="abnf">Timing-Allow-Origin</pre></dd>
+<dt>Applicable protocol:</dt><dd>http</dd>
+<dt>Status:</dt><dd>provisional</dd>
+<dt>Author/Change controller:</dt><dd><a href="https://www.w3.org/">W3C</a></dd>
+<dt>Specification document:</dt><dd><a href="#sec-timing-allow-origin"></a></dd>
+</dl>
 </section>
 <section id="attribute-descriptions">
  <h3>Resource Timing Attributes</h3>
--- a/index.html
+++ b/index.html
@@ -369,7 +369,7 @@ interface:</p>
 "<code>resource</code>".</dd>
 <dt><dfn>startTime</dfn></dt>
 <dd><p data-dfn-for="PerformanceResourceTiming">On getting, the <a>startTime</a> attribute
-returns the result of calling <a>convert fetch timestamp</a> given <a>this</a>'s
+returns the result of calling <a>convert fetch timestamp</a> for <a>this</a>'s
 <a data-for="PerformanceResourceTiming">timing info</a>'s <a for="fetch timing info">
 start time</a> and the <a>relevant global object</a> for <a>this</a>.
 </dl>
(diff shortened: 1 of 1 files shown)


noamr:
@yoavweiss @annevk: this is how the FETCH integration would look like from the RT side.

noamr:
Some explanation:

- RT receives a `response` and a `global object`, and is responsible for queuing, dealing with the performance buffers, and coarsening/adjusting the response timestamps to the global time origin. The timestamps themselves are defined in FETCH.
- FETCH is responsible for TAO - RT calls an algorithm `retrieve timing info` for response, which sanitizes the timing info returned to account for TAO.
- Everything is done in getters to an associated `fetch timing info` - this allows for cases like navigation timing (or future cases of ResourceTiming) where the timestamps for an entry might be updated after the entry is created.

index.html
```diff
@@ -945,292 +651,34 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-<li><em>Final record:</em> If <a
-data-link-for="PerformanceResourceTiming">responseEnd</a>
-is not set, set it to the current time. <dfn data-lt=
-"step-final-record">Record the difference between
-<a>responseEnd</a> and <a>startTime</a> in
-<a>duration</a></dfn>.</li>
-<li>If the fetched resource results in an HTTP redirect or
-<a data-cite="HTML#or-equivalent" data-lt=
-'HTTP response codes equivalence'>equivalent</a>, then
+<li><a>Queue a global task</a> on the <a>networking task source</a> with
```
annevk, line 660: Why do we need a task here? Aren't we already on a task so this can be done synchronously?

annevk (commented):
I really like how this becomes an API just returning values from a model rather than a bunch of prose you have to mentally parse into a model.

One thing I realized is that with our current approach to timers we leave the raw time vulnerable to Spectre attacks. Implementations can probably invert it, but hmm.

index.html
```diff
@@ -254,7 +254,7 @@ <h3>Resources Included in the <a>PerformanceResourceTiming</a>
 <a data-cite=
 "PERFORMANCE-TIMELINE-2#performance-timeline">Performance
 Timeline</a>, unless excluded from the timeline as part of the <a
-href="#processing-model">processing model</a>.
+href="FETCH#concept-fetch">fetching process</a>.
```
yoavweiss, line 257: Are all exclusions happening in Fetch? e.g. where would we be dropping exposure of cross-origin no-cors fetched-by-CSS resources?

yoavweiss (commented):
> One thing I realized is that with our current approach to timers we leave the raw time vulnerable to Spectre attacks. Implementations can probably invert it, but hmm.

Do you mean that we leave the raw timers in the renderer's memory, and a Spectre attacker can use low resolution timers to steal high-resolution timers and amplify the attacks?

index.html
```diff
@@ -61,7 +61,7 @@
     subjectPrefix: "[ResourceTiming]",
```
yoavweiss, line 61: I think you can also add yourself as editor, if you're interested

index.html
```diff
@@ -411,385 +399,103 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
...
-<li>zero, otherwise.</li>
+<dfn>initiatorType</dfn> attribute returns the <a data-for="PerformanceResourceTiming">initiator
+type</a> for <a>this</a>.
+<p data-dfn-for="PerformanceResourceTiming">On getting, the <dfn>duration</dfn> attribute
+returns the result of calling <a>convert fetch timestamp</a> for <a>this</a>'s
+<a data-for="PerformanceResourceTiming">timing info</a>'s <a for="fetch timing info">
+response end time</a> minus <a>this</a>'s
+<p data-dfn-for="PerformanceResourceTiming">On getting, the <dfn>workerStart</dfn> attribute
+returns the result of calling <a>convert fetch timestamp</a> for <a>this</a>'s
+<a data-for="PerformanceResourceTiming">timing info</a>'s <a for="fetch timing info">
```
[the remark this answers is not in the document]
annevk: The modern style would be:

> The X getter steps are to return the result of calling ...

For the latter part, we'd have to do the conversion earlier somehow.

index.html
noamr: Yes, all exclusions happen in FETCH. The `timing info` is attached to a response only if we're passed the CORS checks, and at that point it's also cleaned up for TAO.

index.html
noamr: Not necessarily. Someone needs to queue a task on the global object. We can have FETCH queue this task on the `taskDestination` instead, but there is no guarantee that it's the global object.
Since we anyway need a global object for the time origin, I thought making this re-entrant and queue it here would be right, but open to do this in a different way.

mikewest:
> One thing I realized is that with our current approach to timers we leave the raw time vulnerable to Spectre attacks.

This seems true, but I'm not sure I see the risk. If the attacker has a timer that's good enough to use Spectre-like attacks to access the raw timer, then they don't _need_ the raw timer in the first place.

annevk:
In theory raw timers could be used for cross-process attacks. In practice this is probably infeasible, but it's always nice if the model is the best it can be. There might also be some fingerprinting risks. If I help out figuring out how we can do this, would you be okay with it? (It seems to me that fetch has access to the cross-origin isolated capability through the client so we could use that to obtain a time in the granularity we need, although this will need a new hook.)

mikewest:
I'm not sure it's essential, but if there's a reasonable way to spec things such that we don't need to store the information in a renderer, wonderful! I'd certainly support that, and as you say, it does seem possible for resource timing.

noamr:
> In theory raw timers could be used for cross-process attacks. In practice this is probably infeasible, but it's always nice if the model is the best it can be. There might also be some fingerprinting risks. If I help out figuring out how we can do this, would you be okay with it? (It seems to me that fetch has access to the cross-origin isolated capability through the client so we could use that to obtain a time in the granularity we need, although this will need a new hook.)

You mean, coarsen the timestamps earlier and not when they're accessed through the getter?

annevk:
Yeah, exactly.

index.html
```diff
@@ -945,292 +656,31 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-navigation and define subsequent timing attributes in terms of a
-monotonic clock measuring time elapsed from the beginning of the
-navigation.</p>
-</section>
+</ol>
+<p>To <dfn>convert fetch timestamp</dfn> given <a>DOMHighResTimeStamp</a> |ts| and
+<a>global object</a> |global|, do the following:
+<ol>
+<li>If |ts| is zero, return zero.
+<li>Otherwise, return the <a>relative high resolution time</a> given |ts| and |global|.
```
annevk, line 682: This now coarsens |ts| twice. I'm not sure if that's a problem?

noamr: It's not. But I think it doesn't need to be re-coarsened. Let me check if it can be improved

annevk:
`nextHopProtocol` should also become a `ByteString`.

index.html
yoavweiss: Yeah, coarsening twice is not ideal. Maybe we can expose the diff algorithm from the timestamps to the timeorigin as a separate algorithm?

noamr: Why not. once https://github.com/w3c/hr-time/pull/114 is in I'll amend this one.

index.html
```diff
@@ -944,293 +646,49 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-is fired</a> at the active worker record the time as
-<a>workerStart</a>. Otherwise, if there is no matching
-<a data-cite="service-workers-1#dfn-service-worker-registration">service
-worker registration</a>, set <a>workerStart</a> value to zero.</li>
-<li>If the resource fails the <a>timing allow check</a> algorithm, the user
-agent MUST run the following substeps:
-  <ol>
-    <li>If the resource's <a>Request</a>'s
-    <a data-cite="Fetch#concept-request-destination">destination</a> equals to
-    "document", set <a>workerStart</a> to zero.
```
yoavweiss, line 1092: Here I believe the Fetch integration resulted in setting workerStart to zero for all destinations. I think it's cleaner, but it's a behavior change so we want to make sure implementations align with that.

We should file bugs and add tests on that front.

yoavweiss: /cc @wanderview in case he has opinions given [past discussion](https://bugs.chromium.org/p/chromium/issues/detail?id=1026404#c13)

index.html
```diff
@@ -944,293 +646,49 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-of fetchStart.</li>
-<li>Let redirectEnd be the value of responseEnd.</li>
-<li>Set all the attributes in the <a>PerformanceResourceTiming</a>
-object to 0 except <a>startTime</a>, <a>redirectStart</a>,
-<a>redirectEnd</a>, and <a>initiatorType</a>.</li>
-<li>Return to the step labeled <a href="#dfn-step-fetch-start">Fetch
-start</a> with the redirected resource.</li>
-</ol>
-</li>
+<li>Create a <a>PerformanceResourceTiming</a> object, with its
```
yoavweiss, line 672: Let's create it with `|global|`'s Realm.

index.html
```diff
@@ -944,293 +646,49 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-monotonic clock measuring time elapsed from the beginning of the
-navigation.</p>
-</section>
+</ol>
+<p>To <dfn>convert fetch timestamp</dfn> given {{DOMHighResTimeStamp}} |ts| and
+<a>global object</a> |global|, do the following:
+<ol>
+<li>If |ts| is zero, return zero.
+<li>Otherwise, return the <a data-cite="HR-TIME#relative-high-resolution-coarse-time">relative
+high resolution coarse time</a> andgiven |ts| and |global|.
```
yoavweiss, line 690: "andgiven" should be "given"?

index.html
```diff
@@ -1240,18 +698,17 @@ <h2>Privacy and Security</h2>
 <a>PerformanceResourceTiming</a> interface, the <a data-cite=
 "HTML#same-origin">same origin</a> policy is enforced by default
 and certain attributes are set to zero, as described in <a href=
-"#sec-cross-origin-resources"></a>. Resource providers can
+[=/HTTP fetch=]. Resource providers can
 explicitly allow all timing information to be collected for a
-resource by adding the <a>Timing-Allow-Origin</a> HTTP response
+resource by adding the [<a data-cite="FETCH#tao-check">Timing-Allow-Origin</a>] HTTP response
```
yoavweiss, line 703: Do you still need the data-cite here and below?

index.html
```diff
@@ -944,293 +646,49 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-<img src="timestamp-diagram.svg" alt="Resource Timing attributes"
-style='margin-top: 1em'></figure>
-
-<p>For each resource whose <a>Request</a> has a non-null <a>client</a>, perform
-the following steps:</p>
-<ol data-link-for="PerformanceResourceTiming">
-<li>If the resource is <a data-cite="Fetch#concept-fetch">fetched</a> by a
-cross-origin stylesheet which was fetched with <code>no-cors</code> policy,
-abort the remaining steps.
-<p class="issue">Above cross-origin exclusion should be defined via
```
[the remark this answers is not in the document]
noamr: Will do

index.html
```diff
@@ -944,293 +646,49 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-implementation passes the <a
-href="https://wpt.fyi/results/resource-timing/no-entries-for-cross-origin-css-fetched.sub.html?label=experimental"
->related test</a>.</p>
-</li>
-<li>If the resource's <a>Request</a>'s
-<a data-cite="Fetch#concept-request-destination">destination</a> equals to
-"document", and the <a>Request</a> was not triggered by
-<a data-cite="HTML#process-the-iframe-attributes">process the iframe attributes</a> or
-<a data-cite="HTML#process-the-frame-attributes">process the frame attributes</a>,
-abort the remaining steps.</li>
```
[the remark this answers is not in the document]
noamr: Fetch doesn't handle it, the HTML spec should avoid reporting to RT in those cases.

index.html
```diff
@@ -944,293 +646,49 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-<a>requestStart</a>, <a>responseStart</a>,
-<a>secureConnectionStart</a>, <a>transferSize</a>,
-<a>encodedBodySize</a> and <a>decodedBodySize</a>.</p>
-<p>Server-side applications may return the
-<a>Timing-Allow-Origin</a> HTTP response header to allow the User
-Agent to fully expose, to the document origin(s) specified, the
-values of attributes that would have been zero due to the
-<a>cross-origin</a> restrictions previously specified in this
-section.</p>
-<section id="sec-timing-allow-origin">
```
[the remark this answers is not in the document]
noamr: The TAO header is defined inside FETCH. https://fetch.spec.whatwg.org/#concept-tao-check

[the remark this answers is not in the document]
yoavweiss: We still need ABNF definition and registering it with IANA, which currently RT covers. I think it's fine to keep it here.

index.html
```diff
@@ -946,291 +648,105 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
+ <p>Server-side applications may return the
+ <a data-cite="FETCH#concept-tao-check">Timing-Allow-Origin</a>
+ HTTP response header to allow the User
+ Agent to fully expose, to the document origin(s) specified, the
+ values of attributes that would have been zero due to the
+ <a>cross-origin</a> restrictions previously specified in this
+ section.</p>
+<section id="sec-timing-allow-origin"> 
+ <h4><code>Timing-Allow-Origin</code> Response Header</h4>
+<p class="note">This section is non-normative.</p>
```
yoavweiss, line 673: This section should be normative.

index.html
```diff
@@ -946,291 +648,105 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-"FETCH#http-new-header-syntax">wildcard</a> )
-    </pre>
-<p>The sender MAY generate multiple <a>Timing-Allow-Origin</a>
-header fields. The recipient MAY combine multiple
-<a>Timing-Allow-Origin</a> header fields by appending each
-subsequent field value to the combined field value in order,
-separated by a comma.</p>
-<p>The <dfn>timing allow check</dfn> algorithm, which checks
+ <h3>Cross-origin Resources</h3>
+ <p class="note" data-dfn-for="PerformanceResourceTiming">As detailed in [=fetch=],
```
yoavweiss, line 652: Nit: s/fetch/Fetch/ ?

index.html
```diff
@@ -946,291 +648,105 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
+ cross-origin resources are included as <a>PerformanceResourceTiming</a> objects in the
+ <a data-cite="PERFORMANCE-TIMELINE-2#performance-timeline">Performance
+ Timeline</a>. If the <a href="FETCH#concept-tao-check">timing allow check</a> algorithm
+ fails for a resource, these attributes of its <a>PerformanceResourceTiming</a>
```
yoavweiss, line 656: Nit: s/these/the following/

index.html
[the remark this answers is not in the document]
yoavweiss: Let's open an issue to ensure we explicitly note at these callers that they don't report RT entries to their parent in those cases.

noamr:
Opened HTML issue https://github.com/whatwg/html/issues/6542 to track specific elements that should/shouldn't be integrated.

index.html
```diff
@@ -254,7 +259,7 @@ <h3>Resources Included in the <a>PerformanceResourceTiming</a>
 <a data-cite=
 "PERFORMANCE-TIMELINE-2#performance-timeline">Performance
 Timeline</a>, unless excluded from the timeline as part of the <a
-href="#processing-model">processing model</a>.
+href="FETCH#concept-fetch">fetching process</a>.
```
npm1, line 262: This link does not seem to work in the Preview. Is this a Preview bug or does it need to be fixed?

index.html
```diff
@@ -362,32 +367,26 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
...
 interface:</p>
 <dl data-link-for="PerformanceResourceTiming">
-<dt><dfn>name</dfn></dt>
-<dd>This attribute MUST return the <a data-cite=
-"HTML#resolve-a-url">resolved URL</a> of the requested resource.
-This attribute MUST NOT change even if the <a data-cite=
-"FETCH#concept-fetch">fetch</a> redirected to a different URL.</dd>
+<dt><dfn>name</dfn>
+<dd>The <a>name</a> getter steps are to return the
+<a data-for="PerformanceResourceTiming">requested URL</a>
```
npm1, line 372: The requested URL of who? I think you need to use `this`. Also, nit: no period at end of sentence.

index.html
```diff
@@ -362,32 +367,26 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
...
-[[HR-TIME-2]] with the time immediately before the user agent
-starts to queue the resource for <a data-cite="FETCH#concept-fetch"
-data-lt='fetch'>fetching</a>. If there are HTTP redirects or
-<a data-cite="HTML#or-equivalent">equivalent</a> when fetching the
-resource, and if the <a>timing allow check</a> algorithm passes, this
-attribute MUST return the same value as <a>redirectStart</a>.
-Otherwise, this attribute MUST return the same value as
-<a>fetchStart</a>.</dd>
+<dd><p data-dfn-for="PerformanceResourceTiming">The <a>startTime</a> getter
+steps are to <a>convert fetch timestamp</a> for <a>this</a>'
```
npm1, line 377: nit: sadly it would be `this's`, here and throughout.

index.html
```diff
@@ -362,32 +367,26 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
...
+<a data-for="PerformanceResourceTiming">timing info</a>'s
+[=fetch timing info/start time=] and <a>this</a>'
+<a>relevant global object</a>.
 <dt><dfn>duration</dfn></dt>
-<dd>The <a>duration</a> attribute MUST return a {{DOMHighResTimeStamp}} equal
-to the difference between <a>responseEnd</a> and <a>startTime</a>,
-respectively.</dd>
+<dd>
+<p data-dfn-for="PerformanceResourceTiming">The <a>duration</a> getter
+steps are to <a>convert fetch timestamp</a> for <a>this</a>'
```
npm1, line 384: `convert fetch timestamp` receives two parameters, so you need to add the relevant global in both uses here. Alternatively, maybe it always uses the context object's global object so you could get rid of the second parameter there?

index.html
```diff
@@ -362,32 +367,26 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
 "PERFORMANCE-TIMELINE-2#the-performanceentry-interface">PerformanceEntry</dfn></code>
 interface:</p>
 <dl data-link-for="PerformanceResourceTiming">
-<dt><dfn>name</dfn></dt>
-<dd>This attribute MUST return the <a data-cite=
-"HTML#resolve-a-url">resolved URL</a> of the requested resource.
-This attribute MUST NOT change even if the <a data-cite=
-"FETCH#concept-fetch">fetch</a> redirected to a different URL.</dd>
+<dt><dfn>name</dfn>
```
npm1, line 370: It seems more logical to me to place this after the IDL and after announcing the associated values to the object (some of them used in here). So basically I suggest moving this to right before JSON

index.html
```diff
@@ -411,385 +410,88 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
...
-<a data-cite="RFC7231#section-3.1.2.1">content-codings</a>, if the
-resource is retrieved from <a data-cite=
-"HTML#relevant-application-cache">relevant application caches</a>
-or from local resources.</li>
-<li>zero, otherwise.</li>
+<p data-dfn-for="PerformanceResourceTiming">When <dfn>toJSON</dfn>
+is called, run [[WEBIDL]]'s <a data-cite=
+"WEBIDL#default-tojson-operation">default toJSON operation</a>.</p>
+<p data-dfn-for="PerformanceResourceTiming">
+<dfn>initiatorType</dfn> getter steps would be to return the
```
npm1, line 426: nit: "The initiatorType getter steps are to return", for consistency with the rest?

index.html
```diff
@@ -411,385 +410,88 @@ <h3>The <dfn>PerformanceResourceTiming</dfn> Interface</h3>
...
+<p data-dfn-for="PerformanceResourceTiming">The <dfn>fetchStart</dfn> getter steps are to
+<a>convert fetch timestamp</a> for <a>this</a>' <a data-for="PerformanceResourceTiming">timing
+info</a>'s [=fetch timing info/post-redirect start time=] and the <a>relevant global object</a>
+for <a>this</a>. See [=/HTTP fetch=] for more info.
+<p data-for="PerformanceResourceTiming">The <dfn>domainLookupStart</dfn> getter steps are to
+<a>convert fetch timestamp</a> for <a>this</a>' <a data-for="PerformanceResourceTiming">timing
+info</a>'s [=fetch timing info/final connection timing info=]'s
+[=connection timing info/domain lookup start time=] and the <a>relevant global object</a> for
+<a>this</a>.
+See <a data-cite="FETCH#concept-recording-connection-timing-info">Recording connection timing
```
npm1, line 450: I guess this link will only work after landing the FETCH part? Not currently linking to anything

index.html
```diff
@@ -1239,19 +754,17 @@ <h2>Privacy and Security</h2>
 requested that resource. To limit the access to the
 <a>PerformanceResourceTiming</a> interface, the <a data-cite=
 "HTML#same-origin">same origin</a> policy is enforced by default
-and certain attributes are set to zero, as described in <a href=
-"#sec-cross-origin-resources"></a>. Resource providers can
+and certain attributes are set to zero, as described in [=/HTTP fetch=]. Resource providers can
 explicitly allow all timing information to be collected for a
-resource by adding the <a>Timing-Allow-Origin</a> HTTP response
+resource by adding the [=request/timing allow failed flag|Timing Allow Origin=] HTTP response
```
npm1, line 759: This is the wrong link. You're talking about a header but linking to a flag.

index.html
```diff
@@ -1239,19 +754,17 @@ <h2>Privacy and Security</h2>
...
 header, which specifies the domains that are allowed to access the
 timing information.</p>
 <p>Statistical fingerprinting is a privacy concern where a
 malicious web site may determine whether a user has visited a
 third-party web site by measuring the timing of cache hits and
 misses of resources in the third-party web site. Though the
 <a>PerformanceResourceTiming</a> interface gives timing information
-for resources in a document, the <a href=
-"#sec-cross-origin-resources">cross-origin restrictions</a> prevent
+for resources in a document, the <a data-cite="FETCH#">cross-origin restrictions</a> prevent
```
npm1, line 767: ditto does this link work?

npm1 (commented):
Looks great! Pointing out some nits

index.html
noamr: fixed

index.html
noamr: fixed

index.html
noamr: fixed

index.html
noamr: fixed

index.html
noamr: fixed

index.html
noamr: fixed

index.html
noamr: fixed

index.html
noamr: fixed

index.html
noamr: fixed

index.html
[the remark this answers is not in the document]
noamr: I've put that part back.

index.html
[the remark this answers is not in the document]
noamr: Done

index.html
```diff
@@ -946,291 +641,95 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-<dt>Status:</dt><dd>provisional</dd>
-<dt>Author/Change controller:</dt><dd><a href="https://www.w3.org/">W3C</a></dd>
-<dt>Specification document:</dt><dd><a href="#sec-timing-allow-origin"></a></dd>
-</dl>
-</section>
+ <h3>Cross-origin Resources</h3>
+ <p class="note" data-dfn-for="PerformanceResourceTiming">As detailed in [=Fetch=],
+ cross-origin resources are included as <a>PerformanceResourceTiming</a> objects in the
+ <a data-cite="PERFORMANCE-TIMELINE-2#performance-timeline">Performance
+ Timeline</a>. If the <a href="FETCH#concept-tao-check">timing allow check</a> algorithm
```
npm1, line 648: Link doesn't seem to work

index.html
```diff
@@ -686,16 +679,8 @@ <h4><code>Timing-Allow-Origin</code> Response Header</h4>
...
-whether a resource's timing information can be shared with the
- <a>current document</a>, is as follows:</p>
- <ol>
-     <li>Let <var>response</var> be the resource's <a data-cite=
-       "FETCH#concept-response">Response</a>.</li>
-     <li>Return <var>response</var>'s <a data-cite=
-       "FETCH#concept-response-timing-allow-passed">
-       timing allow passed flag</a>.</li>
- </ol>
+ <p>The <a>Timing-Allow-Origin</a> check is performed as part of
```
npm1, line 682: TAO is a header, not a check, so you may want to word this more precisely (the headers are processed in fetch to compute the attributes accordingly?)

npm1 (approved):
Looks good to me

index.html
noamr: fixed

index.html
noamr: fixed

index.html
```diff
@@ -946,291 +641,95 @@ <h3>Extensions to the <code>Performance</code> Interface</h3>
...
-      "FETCH#concept-response-timing-allow-passed">
-      timing allow passed flag</a>.</li>
-</ol>
-<p class=note>The Timing-Allow-Origin header may arrive as part of a cached
-response. In case of cache revalidation, according to
-<a href="https://tools.ietf.org/html/rfc7234#section-4.3.4">RFC 7234</a>,
-the header's value may come from the revalidation response, or if not present
-there, from the original cached resource.</p>
-</section>
-<section id="sec-iana-considerations">
```
yoavweiss, line 1002: Can you bring back the IANA considerations section?

index.html
yoavweiss: Can you open a Resource Timing issue on the fact that we need better tests here to verify the new behavior? (as well as reporting that as an implementation issues)

yoavweiss (approved):
LGTM % nits and comments

index.html
noamr: Added https://github.com/w3c/resource-timing/issues/264

PR #266, opened 2021-04-13
Expose transferSize, encodedBodySize, decodedBodySize

noamr:
transferSize is based on the cache mode of the resource.
Closes #238

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/pull/266.html" title="Last updated on Apr 15, 2021, 6:02 AM UTC (27c8bda)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/266/7c535a9...27c8bda.html" title="Last updated on Apr 15, 2021, 6:02 AM UTC (27c8bda)">Diff</a>

PR #272, opened 2021-05-07
Add a revalidation cache state and adapt transferSize to it

yoavweiss:
Complementary PR to https://github.com/whatwg/fetch/pull/1232 

While [implementing](https://chromium-review.googlesource.com/c/chromium/src/+/2878693) the [recent Resource Timing transferSize changes](https://github.com/w3c/resource-timing/pull/266), I realized that the current spec is not sufficient to implement what was discussed in the WG, and more importantly, insufficient to maintain backwards compatibility with current transferSize values.

Essentially, we want transferSize to return different values for the cases of local caching (in which case it needs to return 0), full fetch from the network (in which case it needs to return the encoded body size + a constant) as well as the successful revalidation case (where it needs to return just the constant representing the headers used to revalidate the response).

***
<a href="https://pr-preview.s3.amazonaws.com/yoavweiss/resource-timing/pull/272.html" title="Last updated on Jul 12, 2021, 1:47 PM UTC (5cdc106)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/272/3ab43b0...yoavweiss:5cdc106.html" title="Last updated on Jul 12, 2021, 1:47 PM UTC (5cdc106)">Diff</a>


yoavweiss:
@achristensen07 - I missed the fact that this PR hasn't landed yet in yesterday's call. Can you maybe take a look?

nicjansma (approved):

PR #283, opened 2021-07-13
Update and rename pr-push.yml to auto-publish.yml

marcoscaceres:
Renamed the action, as we have scripts at the w3c that look for the file name "auto-publish.yml"

### DIFF
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -1,4 +1,4 @@
-# .github/workflows/pr-push.yml
+# .github/workflows/auto-push.yml
 name: CI
 on:
   pull_request: {}


.github/workflows/auto-publish.yml
```diff
@@ -1,4 +1,4 @@
-# .github/workflows/pr-push.yml
+# .github/workflows/auto-push.yml
```
yoavweiss, line 1: You also need to rename the yml file, right?

yoavweiss: Also, `auto-publish.yml`?

plehegar: yes, and this PR does that.

plehegar (approved):

PR #284, opened 2021-07-13
chore: add tidy workflow

### LINKED ISSUE #282: Running tidy locally results in churn
I tried to run tidy locally after my latest PR, but it resulted in a lot of unrelated changes.
Running tidy effectively means that pending PRs have to be merged back manually, so ideally, we'd have a way to run it in a consistent way as part of CI (so that tidy changes only apply to new prose and not to existing one).

/cc @marcoscaceres @caribouW3 @nicjansma

marcoscaceres:
closes #282

### DIFF
--- a/.github/workflows/tidy.yml
+++ b/.github/workflows/tidy.yml
@@ -1,0 +1,20 @@
+name: Tidy document
+on:
+  workflow_dispatch: {}
+  push:
+    branches:
+      - gh-pages
+
+jobs:
+  tidy:
+    name: Tidy up
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: brew install tidy-html5
+      - run: tidy -config tidyconfig.txt -o index.html index.html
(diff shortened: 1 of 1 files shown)


.github/workflows/tidy.yml
```diff
@@ -0,0 +1,20 @@
...
+
+jobs:
+  tidy:
+    name: Tidy up
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: brew install tidy-html5
+      - run: tidy -config tidyconfig.txt -o index.html index.html
+      - uses: peter-evans/create-pull-request@v3
```
yoavweiss, line 16: Is this what we typically do? Would it make sense to fork that action somewhere "official"?

marcoscaceres: Yeah, that's typical. Open source, y'all. We don't share anything particularly sensitive here, so it's all good.

yoavweiss (approved):

PR #315, opened 2022-01-14
Override spec status for auto-publishing

yoavweiss:
When looking into why RT is not auto-published, it seems like the cause is the spec status that's an editor's draft. NT seems to get over the same issue by overriding the spec status, so trying the same here


caribouW3:
I think it was originally intended to be excluded from auto-publishing until we got a new CR, but since the transition is stuck on the privacy issues (https://github.com/w3c/transitions/issues/337), I think we can republish WDs.

caribouW3 (approved):

PR #323, opened 2022-04-10
Illustrate in diagram that timestamps are not necessarily consecutive

### LINKED ISSUE #164: elaborate on `stalled` or `blocked` stage in specs

noamr:
Closes #164


yoavweiss:
Thanks!
I'm not sure "stall" is the best word to describe the delays that can be incurred from the browser exceeding the DNS, TCP/QUIC connection and in-slight request limits. Maybe a "potential delay" bubble below, with arrows pointing to before the DNS, TCP/QUIC and request squares?

noamr:
> Thanks! I'm not sure "stall" is the best word to describe the delays that can be incurred from the browser exceeding the DNS, TCP/QUIC connection and in-slight request limits. Maybe a "potential delay" bubble below, with arrows pointing to before the DNS, TCP/QUIC and request squares?

What's the word used for this in DevTools? I want to use the same vocabulary

yoavweiss:
Chrome DevTools indeed uses the word "stalled" for the connection parts, but doesn't talk about other types of delays. IMO, it's inaccurate, but there's no point in bikeshedding the diagram, as it's easy to change it in the future if complaints come in.

What about changing it to a single bubble pointing it at the 3 potential locations?

yoavweiss (approved):
LGTM! Looks great :)

Issue #303, opened 2021-10-26
Reflect preload information (in ResourceTiming?)


noamr:
Right now it's impossible to know whether a resource was used as part of a preload.
Preloading the resource would generate the RT entry, and in some cases it would be reused without a new fetch (e.g. if it's in the image cache), and in some cases there would be a new fetch that would go to cache (with `transferSize` 0).

Suggesting to be explicit about it, and to say in a subsequent fetch whether it was served from network, service-worker, HTTP cache, or preload.

noamr:
Actually based on the newer developments in the preload PR, I tend to agree with either having some "unused preload" flag (maybe something like `readyState` on the `link` element?), and/or having a `preloadConsumedTime` attribute on the RT entry.

noamr:
Proposal:
- A `preload` link, once loaded, creates an RT entry immediately, but doesn't queue it to the PerformanceObservers.
- Once the preload is consumed, it marks a `preloadConsumedTime` timestamp on the same entry, and the entry gets queued (maybe it somehow register the consumer `initiatorType`).
- The `duration` would be based on `responseEnd` like today, and the `preloadConsumedTime` would be (most likely) after that, though in some cases it can be before (if the resource is used before it's fully loaded).

This is in line with navigation timing, where the entry is created when we know about it, but only gets queued when its final attribute is set.

It also allows us to report unused preloads in RUM without adding a timeout - a RUM library or a document can check on its own whether we have preloaded links that don't match RT entries with corresponding `preloadConsumedTime`.

WDYT, @andydavies @yoavweiss

yoavweiss:
I like it!
/cc @npm1 @philipwalton @pmeenan for their opinions.

npm1:
The proposal seems reasonable to me, but it's unclear from the discussion in the thread what the current behavior is and why we need to change it. Can this be documented here?

noamr:
> The proposal seems reasonable to me, but it's unclear from the discussion in the thread what the current behavior is and why we need to change it. Can this be documented here?

The current behavior doesn't mark the time the preload was consumed, and you'd need to guess whether it's consumed based on the existence of another RT entry with the same URL.

nicjansma:
@noamr I like your proposal.  Two potential challenges though:

* For HTTP Header-based Preloads, RUM doesn't know what those headers were (or if there were any).  So RUM wouldn't be able to cross-check with the RT entries we saw to know which header-based Preloads weren't used.
[...]

Issue #262, opened 2021-04-01
Render blocking status of resources


yoavweiss:
Right now, it's hard to tell which resources were actually render blocking and which ones were not.
Developers can try to heuristically guess that, based on when those resources were downloaded, but that's hard to do in the lab, never mind in the wild.

Lighthouse [can now](https://github.com/GoogleChrome/lighthouse/issues/2065#issuecomment-768162010) get rid of those heuristics, as Chromium provides a direct signal for render blocking resources (at least for scripts and styles).

It'd be good if RUM had similar capabilities, that would enable similar analysis by RUM providers:
* Which resources were downloaded before FCP but were not render blocking? (and hence, could've been delayed)
* Which resources were render blocking, but weren't discovered early enough? (and hence, could've benefited from being preloaded)

And probably more...

yoavweiss:
^^ @crockercliff

ArthurSonzogni:
Hi @abinpaul1

We were doing the security/privacy review of:
https://github.com/abinpaul1/resource-timing/blob/render-blocking-status-explainer/Explainer/Render_Blocking_Status.md

It would help if you could add a section explaining why this isn't revealing new cross-origin information. You could also fill https://github.com/w3ctag/security-questionnaire

From my reading of the spec, it should be not reveal cross-origin information. The resources will be the ones fetched from <link>,<script>,<style> element with blocking="render" attribute. The information is already defined in the document.
What about resources nested inside an <iframe> or a <script>?

abinpaul1:
Hi @ArthurSonzogni,
I've updated the explainer with a section on Privacy/Security considerations along with the questionnaire. 

> What about resources nested inside an <iframe> or a <script>?

- With respect to ResourceTiming, each frame only reports on the resources fetched by itself, and not any child IFRAMEs. 
- For resources (scripts) imported by scripts, entries are created for them and their `renderBlockingStatus` would be `non-blocking`. Currently based on the spec, script resources not inserted by parser are only render blocking if blocking=render attribute is present.

Does this help with your query?

ArthurSonzogni:
> Does this help with your query?

It does! Thanks you!

yoavweiss:
@abinpaul1 - I think we can close this as "fixed", right?

PR #327, opened 2022-06-02
Add Render Blocking Status to PerformanceResourceTiming

abinpaul1:
Introducing a render blocking status field. (https://github.com/w3c/resource-timing/issues/262)
Fetch changes : https://github.com/whatwg/fetch/pull/1449
~~HTML changes: https://github.com/whatwg/html/pull/7979~~

Explainer : https://github.com/abinpaul1/resource-timing/blob/render-blocking-status-explainer/Explainer/Render_Blocking_Status.md

***
<a href="https://pr-preview.s3.amazonaws.com/abinpaul1/resource-timing/pull/327.html" title="Last updated on Aug 10, 2022, 6:11 PM UTC (41acb22)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/327/e71c2ab...abinpaul1:41acb22.html" title="Last updated on Aug 10, 2022, 6:11 PM UTC (41acb22)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -380,7 +380,7 @@
               readonly attribute unsigned long long  transferSize;
               readonly attribute unsigned long long  encodedBodySize;
               readonly attribute unsigned long long  decodedBodySize;
-              readonly attribute DOMString renderBlockingStatus;
+              readonly attribute RenderBlockingStatusType renderBlockingStatus;
               [Default] object toJSON();
           };
         </pre>
@@ -409,6 +409,11 @@
           A <a>PerformanceResourceTiming</a> has an associated [=response body
           info=] <a data-dfn-for="PerformanceResourceTiming"><dfn>resource
           info</dfn></a>.
+        </p>
+        <p>
+          A <a>PerformanceResourceTiming</a> has an associated
+          {{RenderBlockingStatusType}} <a data-dfn-for=
+          "PerformanceResourceTiming"><dfn>render blocking status</dfn></a>.
         </p>
         <p>
           The <a>PerformanceResourceTiming</a> interface participates in the
@@ -700,10 +705,9 @@
           </li>
         </ol>
         <p data-dfn-for="PerformanceResourceTiming">
-          The <dfn>renderBlockingStatus</dfn> getter steps are to return
-          <code>"blocking"</code> if <a>this</a>'s <a data-for=
-          "PerformanceResourceTiming">timing info</a>'s [=fetch timing info/
-          render-blocking=] is true, else <code>"non-blocking"</code>
+          The <dfn>renderBlockingStatus</dfn> getter steps are to return the
+          <a data-for="PerformanceResourceTiming">render blocking status</a>
+          for <a>this</a>.
         </p>
         <p class='note'>
           A user agent implementing <a>PerformanceResourceTiming</a> would need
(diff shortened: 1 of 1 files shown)


yoavweiss:
Thanks @abinpaul1!!

@caribouW3 - What do we need to appease the IPR bots here? Should Abin join the WG as an invited expert?

caribouW3:
Are we in situation 2. b described here: https://github.com/w3c/Guide/blob/master/process/non-participant-commitment.md
I think that if it's a one-time contribution, a non-participant commitment is enough
(Contributor needs a W3C public account, link it with the GH account, then it's done by submitting a licensing commitment form).

yoavweiss:
We [discussed](https://docs.google.com/document/d/1VKlWpadbyq2iv35JVVUdxFuB2mshEHP_ArPcij8fpk8/edit#heading=h.9gqz90lre5ux) this on the WG call yesterday.

@bdekoz @sefeng211 @achristensen07 - I'd love y'all's thoughts on this

yoavweiss:
Left a [relevant comment](https://github.com/whatwg/fetch/pull/1449#issuecomment-1154949581) on the Fetch PR that would impact this PR as well/

noamr:
Any WPTs for this?

noamr:
Great idea, though I believe this would be confusing with preloads until we resolve https://github.com/w3c/resource-timing/issues/303. If you preload a stylesheet and then load it as render-blocking, it would currently appear as not render-blocking because currently we don't expose preload-consuming requests in resource timing.

abinpaul1:
> Any WPTs for this?

It'll being added over at https://chromium-review.googlesource.com/c/chromium/src/+/3709521
Have added a few cases but its still a WIP

noamr:
It's probably failing because the Fetch PR is not merged yet. I want that merged first and to get some understanding about https://github.com/w3c/resource-timing/pull/327#issuecomment-1159514743 before proceeding

yoavweiss:
I agree that this would be confusing initially, but don't know if we want to block this work on solving preloads. 
@noamr - Would an intermediate state where preload RT enrties are marked as non-blocking be too confusing in your opinion? I believe this is what Chrome is currently doing with traces.

noamr:
> I agree that this would be confusing initially, but don't know if we want to block this work on solving preloads. 
> 
> @noamr - Would an intermediate state where preload RT enrties are marked as non-blocking be too confusing in your opinion? I believe this is what Chrome is currently doing with traces.

Sounds ok, just wanted to get that conversation started (this PR is blocked on the fetch PR)

abinpaul1:
Changed the type of render blocking status from string to `enum` as per TAG suggestion https://github.com/w3ctag/design-reviews/issues/753#issuecomment-1197900871.

index.html
```diff
@@ -698,12 +704,49 @@ <h3>
             </p>
           </li>
         </ol>
+        <p data-dfn-for="PerformanceResourceTiming">
+          The <dfn>renderBlockingStatus</dfn> getter steps are to return the
+          <a data-for="PerformanceResourceTiming">render-blocking status</a>
+          for <a>this</a>.
```
sefeng211, line 710: So this is rendered as `The renderBlockingStatus getter steps are to return the render-blocking status for this.`

Shouldn't be something like `The renderBlockingStatus getter steps are to return the render-blocking status for this's timing info's render-blocking and the relevant global object for this.`?

abinpaul1: The enum's value is determined [here](https://pr-preview.s3.amazonaws.com/abinpaul1/resource-timing/pull/327.html#sec-render-blocking-status-types) using `timing info's render-blocking` and I thought `getter` would have to just return that value.

Do I have to draft it as `return the render-blocking status for this's timing info's render-blocking and the relevant global object for this` to signify that the `timing info's render-blocking` is used to set the value of enum that is returned?

sefeng211: I see, I missed that enum part. 

I am not too sure...Just because I expect the enum should just define the meaning of each value, not involves in effectively getting the value from `timing info`. 

Maybe I am just picky here and causing unnecessary troubles.

@noamr can confirm..

noamr: Yes it's odd to define a getter in the enum.
The enum should describe the meaning, and the getter steps should be to get the enum value based on the boolean value as @sefeng211 said.

abinpaul1: Cool! I've updated it to do that. Thanks!

index.html
```diff
@@ -698,12 +704,48 @@ <h3>
             </p>
           </li>
         </ol>
+        <p data-dfn-for="PerformanceResourceTiming">
+          The <dfn>renderBlockingStatus</dfn> getter steps are to return
+          <a data-link-for="RenderBlockingStatusType">blocking</a>
+          if <a>this</a>'s <a data-for="PerformanceResourceTiming">timing
+          info</a>'s [=fetch timing info/render-blocking=] is true
+          else <a data-link-for="RenderBlockingStatusType">non-blocking</a>
```
noamr, line 712: ; Otherwise

index.html
```diff
@@ -698,12 +704,48 @@ <h3>
...
+          and the <a>relevant global object</a> for <a>this</a>.
```
noamr, line 713: I don't understand this line, how is the global object used here?

index.html
abinpaul1: Done

index.html
abinpaul1: It's not. Apologies for the confusion. Removed.

noamr (approved):
LGTM

yoavweiss (approved):
LGTM

PR #330, opened 2022-07-05
Explainer for Render Blocking Status

abinpaul1:


caribouW3:
> @caribouW3 - Abin is working on Render Blocking signal in Resource Timing. Can we add him as an Invited Expert?

No objection from my perspective. (https://www.w3.org/groups/wg/webperf/instructions)

Explainers/Render_Blocking_Status.md
```diff
@@ -0,0 +1,159 @@
+# Render Blocking Status
+
+Proposal to add a new field `render blocking status` to PerformanceResourceTiming which is an [enum](https://webidl.spec.whatwg.org/#idl-enums)  highlighting the status of stylesheets and scripts.
+
+## Use cases
+
+RUM providers could then easily determine the render blocking resources without having to rely on complex heurestics to determine the same.
+This would enable analysis for below scenaarios and possibly more.
```
tunetheweb, line 8: Nit:

```suggestion
This would enable analysis for below scenarios and possibly more.
```

Explainers/Render_Blocking_Status.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Render Blocking Status Values
+
+The following values are proposed for the resource blocking status enum
+
+`blocking` - A potentially render blocking resource
+
+`non-blocking` - Non blocking resource
+
+We primarily reuse the value of the [render blocking](https://fetch.spec.whatwg.org/#request-render-blocking) boolean associated with fetch [request](https://fetch.spec.whatwg.org/#concept-request)
```
tunetheweb, line 49:
```suggestion
We primarily reuse the value of the [render-blocking](https://fetch.spec.whatwg.org/#request-render-blocking) boolean associated with fetch [request](https://fetch.spec.whatwg.org/#concept-request)
```

Also the [HTML spec](https://html.spec.whatwg.org/#render-blocking-mechanism) uses "render-blocking" rather than "render blocking", even when not explicitly referring to the `render-blocking` boolean, which seems more correct to me. Should we use that format throughout this doc (which would also avoid errors like this when actually refer to the boolean)?

And, if so, possibly rename the file too to `Render-blocking_Status.md` - or would that mix of dashes and underscores just be confusing?

tunetheweb (changes requested):

Explainers/Render_Blocking_Status.md
abinpaul1: Thanks. I've made the changes everywhere in the doc. It does make sense to allign with the existing terminology in the spec. 

I also think now I should update it in the spec PR (https://github.com/w3c/resource-timing/pull/327) as well, so that that is also in line with the existing terminology. WDYT?

> And, if so, possibly rename the file too to Render-blocking_Status.md - or would that mix of dashes and underscores just be confusing?

This explainer in my repo was linked to various places, so if I change the name those links might become inaccessible. I'll rename it in a follow up PR after this lands?

tunetheweb: Thanks for that. Updating the spec PR similarly seems sensible to me.

tunetheweb (approved):

PR #331, opened 2022-07-06
Fix the assertion in setup the resource timing entry.

jeremyroman:
#272 (a4628756a5337343ae1d9deef915198c2b6a3121) added a third option, `"validated"`, but the assertion was not updated. Consequently anything served from cache after revalidation currently fails this assertion, but should not.


yoavweiss (approved):
LGTM! Thanks for catching that! :)

Issue #203, opened 2019-03-07
Include Content Type


Snugug:
I'd like to be able to use Resource Timing to programmatically get a sense of where my application is spend its performance budget, but unfortunately I cannot because resource timing doesn't include the ability to inspect why kind of resource it is.

I would love to see either the `content-type` header, a [`Request Destination`](https://developer.mozilla.org/en-US/docs/Web/API/RequestDestination) string, or both, included in resource timing to allow for this kind of introspection of the resource.

yoavweiss:
We've discussed adding an `initiator` or some other attribute which maps into `Reqeust.destination`. I believe that would be an improvement over the current `initiatorType`.

nicjansma:
One use case where having access to `Content-Type` would be useful is for analyzing file formats that may be "swapped out" transparently (and the URL doesn't change).

For example, with adoption of WebP and AVIF, some websites will be built with `<picture>` elements where each supported  image format has a different URL.  But there are commercial CDNs out there that transparently swap out the image format based on the User Agent string or Client Hints.  In those cases you may have a `.jpg` URL that actually returns `image/avif`.

If we had access to the `Content-Type` we could know exactly what format was being used.  This can help in cases of understanding format adoption for a website.  For analytics vendors, this could be helpful in segregating user experiences based on the image formats the client decoded.

nicjansma:
We discussed this on the 2021/07/29 W3C WebPerf call, minutes are here: https://docs.google.com/document/d/1O-LPrbJYXZST7AdIZxQIOMFsnikiZpNk6H15mnKWAa8/edit#

Summary:

* There are a couple different potential scenarios/requirements in this ticket from (1) understanding request destination to (2) accessing `Content-Type` to (3) what format the browser interpreted requests at.  Each may have different solutions.
* Some servers may lie about `Content-Type` and that [MIME Sniffing rules](https://mimesniff.spec.whatwg.org/) come into play to determine what the UA may actual read the content as.
    * MIME sniffing rules are not highly spec compliant between browsers, none are over 50% passing and there are a lot of edge-cases so it may not be a reliable spec to use for this
* For the scenario I mentioned before, understanding the "interpreted" Content Type (whatever the browser reads the content as) would be more valuable than the server's raw `Content-Type` header.

alonkochba:
Adding my 2 cents here - we ran into the need to determine our images content-type, specifically for LCP images.
[...]

PR #341, opened 2022-08-22
[draft] Add content-type to resource-timing

abinpaul1:
Introducing Content type. (https://github.com/w3c/resource-timing/issues/203)
Fetch changes : 

Explainer : https://github.com/abinpaul1/resource-timing/blob/explainer-content-type/Explainers/Content-Type.md

### DIFF
--- a/index.html
+++ b/index.html
@@ -380,6 +380,7 @@
               readonly attribute unsigned long long  transferSize;
               readonly attribute unsigned long long  encodedBodySize;
               readonly attribute unsigned long long  decodedBodySize;
+              readonly attribute DOMString contentType;
               [Default] object toJSON();
           };
         </pre>
@@ -408,6 +409,10 @@
           A <a>PerformanceResourceTiming</a> has an associated [=response body
           info=] <a data-dfn-for="PerformanceResourceTiming"><dfn>resource
           info</dfn></a>.
+        </p>
+        <p>
+          A <a>PerformanceResourceTiming</a> has an associated DOMString
+          <a data-dfn-for="PerformanceResourceTiming"><dfn>content type</dfn></a>.
         </p>
         <p>
           The <a>PerformanceResourceTiming</a> interface participates in the
@@ -698,6 +703,11 @@
             </p>
           </li>
         </ol>
+        <p data-dfn-for="PerformanceResourceTiming">
+          The <dfn>contentType</dfn> getter steps are to return
+          <a data-for="PerformanceResourceTiming">content type </a>for
+          <a>this</a>.
+        </p>
         <p class='note'>
           A user agent implementing <a>PerformanceResourceTiming</a> would need
           to include <code>"resource"</code> in
@@ -1062,8 +1072,9 @@
           To <dfn data-export="">mark resource timing</dfn> given a [=/fetch
           timing info=] |timingInfo|, a DOMString |requestedURL|, a DOMString
           |initiatorType| a <a>global object</a> |global|, a string
-          |cacheMode|, and a [=/response body info=] |bodyInfo|, perform the
-          following steps:
+          |cacheMode|, a [=/response body info=] |bodyInfo|, and a
+          <a data-cite="MIMESNIFF#mime-type">MIME type</a> |mimetype|
+          perform the following steps:
         </p>
         <ol>
           <li>Create a <a>PerformanceResourceTiming</a> object |entry| in
@@ -1071,8 +1082,8 @@
           </li>
           <li>
             <a>Setup the resource timing entry</a> for |entry|, given
-            |initiatorType|, |requestedURL|, |timingInfo|, |cacheMode|, and
-            |bodyInfo|.
+            |initiatorType|, |requestedURL|, |timingInfo|, |cacheMode|,
+            |bodyInfo|, and |mimeType|.
           </li>
           <li>
             <a data-cite=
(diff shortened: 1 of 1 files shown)


abinpaul1:
I was trying to spec this out keeping in mind, the points mentioning not to reflect the header directly and only a set of relevant MIME types. 

- It is currently(in this PR) specified to allow any MIME type that matches the groups presented in https://mimesniff.spec.whatwg.org/#mime-type-groups. 
- Taking the case of say a [ZIP-based MIME type](https://mimesniff.spec.whatwg.org/#zip-based-mime-type), its described as any MIME type(https://mimesniff.spec.whatwg.org/#mime-type) whose [subtype](https://mimesniff.spec.whatwg.org/#subtype) ends in "+zip". So from what I understand a content type such as `application/u.26363-43323-42424+zip` would be valid and reflected, which is not desirable. 
- To get around this, AFAIU we would need to define an exhaustive list of allowed MIME types to match against. I don't think we can go with the list that was mentioned in the WG call (https://www.iana.org/assignments/media-types/media-types.xhtml) because a good majority of those do not seem to be relevant for the use cases.
- If we were to keep an exhaustive list, which would be the more appropriate place to keep this? (MIME sniffing/Resource Timing). Such a list would have to updated as well when a new relevant content-type gains traction as well.

yoavweiss:
@abinpaul1 - can we just reflect the MIME type's [essensce](https://mimesniff.spec.whatwg.org/#mime-type-essence)?

abinpaul1:
Yeah, but when we reflect the [essence](https://mimesniff.spec.whatwg.org/#mime-type-essence) wouldn't this be a valid essence : `application/u.26363-43323-42424+zip` ?
Here `application` would be [type](https://mimesniff.spec.whatwg.org/#type) and `u.26363-43323-42424+zip` is the [subtype](https://mimesniff.spec.whatwg.org/#subtype)

yoavweiss:
Yeah, you're right.. Maybe it'd be possible to convince the MIMESNIFF editors to modify that definition, or otherwise monkeypatch it, if ZIP is the only mimetype that can expose such UIDs

/cc @domenic @annevk

abinpaul1:
I don't think `zip` is the only one that can have such UIDs. 
[JSON](https://mimesniff.spec.whatwg.org/#json-mime-type), [XML](https://mimesniff.spec.whatwg.org/#xml-mime-type) MIME types also only requires [subtype](https://mimesniff.spec.whatwg.org/#subtype) to end with `+json`, `+xml`. Morover, [Image](https://mimesniff.spec.whatwg.org/#image-mime-type) and [audio/video](https://mimesniff.spec.whatwg.org/#audio-or-video-mime-type) MIME types does not have any restrictions on [subtype](https://mimesniff.spec.whatwg.org/#subtype) (only requires the [type](https://mimesniff.spec.whatwg.org/#type) to be `image`/`audio`/`video`)

domenic:
Why is the full MIME type string not reflected? I couldn't find any reasoning in the explainer at https://github.com/abinpaul1/resource-timing/blob/explainer-content-type/Explainers/Content-Type.md#content-type-values.

abinpaul1:
During the WG call when this was discussed, a concern was raised regarding reflecting the exact same value returned by server. 
From what I could understand the concern was mainly to protect against having any unique identifier as part of content-type in server response being relflected through resource timing API. @achristensen07, it would be great if you could please expand on the concern.

yoavweiss:
I believe the concerns are related to https://github.com/w3c/server-timing/issues/89 and are planned to be discussed at TPAC.

index.html
```diff
@@ -711,26 +712,22 @@ <h3>
...
         </ol>
         <p data-dfn-for="PerformanceResourceTiming">
-          The <dfn>contentType</dfn> getter steps are to perform the following
-          steps:
+          The <dfn>contentType</dfn> getter steps are to return
+          <a>this</a>'s <a data-for="PerformanceResourceTiming">resource
+          info</a>'s [=response body info/content type=]'s <a data-cite=
+          "MIMESNIFF#mime-type-essence">essence</a>.
+        </p>
+        <p class='note'>
```
yoavweiss, line 720: I think this shouldn't be a note, but a normative part of the prose. Maybe "return this's resource info's content type's essence. The user agent may truncate the value in an [implementation defined](https://infra.spec.whatwg.org/#implementation-defined) way." ?

abinpaul1: Updated

index.html
```diff
@@ -710,6 +711,13 @@ <h3>
             </p>
           </li>
         </ol>
+        <p data-dfn-for="PerformanceResourceTiming">
+          The <dfn>contentType</dfn> getter steps are to return
+          <a>this</a>'s <a data-for="PerformanceResourceTiming">resource
+          info</a>'s [=response body info/content type=]'s <a data-cite=
+          "MIMESNIFF#mime-type-essence">essence</a>. The user agent may
```
yoavweiss, line 718: Can you also link [user agent](https://infra.spec.whatwg.org/#user-agent)?

yoavweiss (approved):
LGTM once the Fetch PR lands

yoavweiss (approved):
LGTM

index.html
```diff
@@ -719,6 +720,16 @@ <h3>
...
+        <p data-dfn-for="PerformanceResourceTiming">
+          The <dfn>contentType</dfn> getter steps are to return
+          <a>this</a>'s <a data-for="PerformanceResourceTiming">resource
+          info</a>'s [=response body info/content type=]'s <a data-cite=
+          "MIMESNIFF#mime-type-essence">essence</a> if
+          <a>this</a>'s <a data-for=
+          "PerformanceResourceTiming">resource info</a>'s
+          [=response body info/content type=] is non-null; otherwise
+          the empty string.
+        </p>
```
annevk, line 732: Rather than just returning the essence, this should do https://mimesniff.spec.whatwg.org/#serializing-a-mime-type, no?

abinpaul1: Yeah, that seems more appropriate. Updated.

RByers:
Is there something blocking this PR from landing?
Context: [blink intent to ship](https://groups.google.com/a/chromium.org/g/blink-dev/c/JzYRpqkpQV8/m/a4xUxNNRAgAJ)

yoavweiss:
This is waiting on the [Fetch PR](https://github.com/whatwg/fetch/pull/1481) to land.

yoavweiss (approved):
LGTM

Issue #332, opened 2022-07-06
Expose delivery type in PerformanceResourceTiming.


jeremyroman:
This is an attempt to address both w3c/navigation-timing#163 and #303 by exposing a `deliveryType` attribute on `PerformanceResourceTiming` (as mentioned in https://github.com/w3c/resource-timing/issues/303#issuecomment-1099078362), specifically the entry that may consume (rather than populate) a cache. This doesn't preclude timing entries being available for the request which populates the cache, if appropriate.

It would be one of the following DOMStrings:

* `"navigational-prefetch"`, which would be used for the cases in w3c/navigation-timing#163. Algorithms which use a prefetched response during navigation would call [mark resource timing](https://w3c.github.io/resource-timing/#dfn-mark-resource-timing) with this delivery type.
* `"preload"`, which would be passed by [consume a preloaded resource](https://html.spec.whatwg.org/multipage/links.html#consume-a-preloaded-resource)
* `"cache"`, which would apply if the cache mode is non-empty (i.e., the resource came from the cache)
* the empty string, if none of the above apply. We could alternatively call this `"network"` or similar, but this has precedent in cache mode, is perhaps less misleading if new delivery types aren't immediately supported, and is false-ish, similar to if the the property is absent.

`"navigational-prefetch"` could be temporarily left out of the spec proper and instead kept as a monkey-patch in the proposals for specifying navigational prefetch, but I've included it for now.

`"preload"` could also land as a separate normative change coordinated with its own HTML and WPT pull requests.

jeremyroman:
@yoavweiss @noamr Trying to unblock these two issues; wdyt?

The delta to this specification is fairly small: https://github.com/w3c/resource-timing/compare/gh-pages...jeremyroman:resource-timing:resource-timing-prefetch?expand=1

yoavweiss:
Right now, consuming a preloaded resource doesn't create an entry. Modifying that would require some discussion, which started in #303. IIUC, your proposal here matched @noamr's "separate entry" [proposal](https://github.com/w3c/resource-timing/issues/303#issuecomment-1099078362). Is that correct?

jeremyroman:
Yes, I propose that consuming a preloaded resource create a separate entry.

noamr:
Yes I think we should have separate entries for each fetch and separate them by something like this proposal. Not sure how to advance towards a consensus on #303 though.

jeremyroman:
[...]

PR #343, opened 2022-08-30
Add delivery type to PerformanceResourceTiming.

jeremyroman:
Addresses #332.

This defines a concept of "delivery type" which currently describes the difference between resources delivered from the HTTP cache and those not (currently only detectable via transferSize). It is expected to be extended to describe resources delivered from other stores, such as the consuming resources from the HTML preload buffer and navigational prefetches.

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/resource-timing/pull/343.html" title="Last updated on Mar 10, 2023, 4:06 PM UTC (78dc701)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/343/98b0e13...jeremyroman:78dc701.html" title="Last updated on Mar 10, 2023, 4:06 PM UTC (78dc701)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1062,7 +1062,8 @@
           <p>
             The user agent MAY still enforce cross-origin restrictions and set
             transferSize, encodedBodySize, and decodedBodySize attributes to
-            zero, even with Timing-Allow-Origin HTTP response header fields.
+            zero, even with Timing-Allow-Origin HTTP response header fields. If
+            it does, it MAY also set deliveryType to "".
           </p>
           <p>
             The <a>Timing-Allow-Origin</a> headers are processed in
--- a/index.html
+++ b/index.html
@@ -364,6 +364,7 @@
           [Exposed=(Window,Worker)]
           interface PerformanceResourceTiming : PerformanceEntry {
               readonly attribute DOMString initiatorType;
+              readonly attribute DOMString deliveryType;
               readonly attribute ByteString nextHopProtocol;
               readonly attribute DOMHighResTimeStamp workerStart;
               readonly attribute DOMHighResTimeStamp redirectStart;
@@ -386,6 +387,11 @@
         <p>
           A <a>PerformanceResourceTiming</a> has an associated DOMString
           <a data-dfn-for="PerformanceResourceTiming"><dfn>initiator
+          type</dfn></a>.
+        </p>
+        <p>
+          A <a>PerformanceResourceTiming</a> has an associated DOMString
+          <a data-dfn-for="PerformanceResourceTiming"><dfn>delivery
           type</dfn></a>.
         </p>
         <p>
@@ -549,6 +555,32 @@
           The setting of `initiatorType` is done at the different places where
           a resource timing entry is reported, such as the [=fetch=] standard.
         </p>
+        <p data-dfn-for="PerformanceResourceTiming">
+          <dfn>deliveryType</dfn> getter steps are to return the <a data-for=
+          "PerformanceResourceTiming">delivery type</a> for <a>this</a>.
+        </p>
+        <div class='note'>
+          <p>
+            `deliveryType` returns one of the following values:
+          </p>
+          <ul>
+            <li>
+              <code>"cache"</code>, if the <a data-for=
+              "PerformanceResourceTiming">cache mode</a> is not the empty
+              string.
+            </li>
+            <li>
+              the empty string <code>""</code>, if none of the above conditions
+              match.
+            </li>
+          </ul>
+
+          <p class='note'>
+            This is expected to be expanded by future updates to this
+            specification, e.g. to describe consuming preloaded resources and
+            prefetched navigation requests.
+          </p>
+        </div>
         <p data-dfn-for="PerformanceResourceTiming">
           The <dfn>workerStart</dfn> getter steps are to <a>convert fetch
           timestamp</a> for <a>this</a>'s <a data-for=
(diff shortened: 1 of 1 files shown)


jeremyroman:
This is missing WPT tests at present. I understand if merging this is blocked on that, but would you mind having a look @noamr ? I expect the tests for this will be relatively straightforward (at least, once we have an implementation in Chromium to avoid test bugs).

index.html
```diff
@@ -1111,6 +1145,12 @@ <h2>
           <li>Set |entry|'s <a data-for="PerformanceResourceTiming">cache
           mode</a> to |cacheMode|.
           </li>
+          <li>If |deliveryTode| is the empty string and |cacheMode| is not,
```
yoavweiss, line 1148: s/Tode/Type/ here and below

jeremyroman: Done.

index.html
```diff
@@ -1111,6 +1145,12 @@ <h2>
           <li>Set |entry|'s <a data-for="PerformanceResourceTiming">cache
           mode</a> to |cacheMode|.
           </li>
+          <li>If |deliverytype| is the empty string and |cacheMode| is not,
```
yoavweiss, line 1148: deliveryType

yoavweiss (approved):
LGTM, but I'd love to run this by the WG and see if there are any objections, as well as get tests in place before this lands.

jeremyroman:
> LGTM, but I'd love to run this by the WG and see if there are any objections, as well as get tests in place before this lands.

Is TPAC a good venue for that? If so are there steps I should take to get it on the agenda?

yoavweiss:
> Is TPAC a good venue for that? If so are there steps I should take to get it on the agenda?

We can probably cover this on [today's call](https://docs.google.com/document/d/10dz_7QM5XCNsGeI63R864lF9gFqlqQD37B4q8Q46LMM/edit?pli=1#heading=h.15anydpwx5wv).

nicjansma:
Summary from W3C WebPerf call:

* Need to figure out cross-origin/TAO/CORS limitations
* Possible objections from Safari, to be discussed further at TPAC in ~2 weeks

yoavweiss:
Thinking out loud regarding the need for restrictions here:
If we don't add any TAO restrictions to this, this doesn't reveal anything about the resource itself, but it can reveal some metadata about the resource. Specifically, it can reveal things about the resource's caching headers. It provides a bit on whether the resource is cachable or not (and potentially, for how long it is cacheable, through continous probing).

While not significant on its own, this could be sufficient to reveal user state in cases where the cachability of the resource varies between e.g. logged-in and non-logged-in users. 

So, I think we want this to be TAO protected.

/cc @arturjanc and @mikewest to tell me if I'm being overly paranoid about this or overly optimistic about this not already being leaked through timing.

jeremyroman:
I think that tracks. I think the strongest counterarguments would be:

1. The risk is substantially mitigated in browsers which partition the cache by at least top-level site (and some also use a secondary key). In such cases the cache state doesn't reveal anything about browsing activity on other top-level sites. I'm not certain but I believe this is now all major browsers.
2. It is likely already deducible from timing heuristics (even without TAO-protected info) in most cases.

I'm not opposed to TAO protection, though, and presumably anyone gathering RUM is also serving their important cross-origin resources with a suitable TAO header.

If that is the resolution, the behavior I propose is that the if `"cache"` would have been the delivery type but the response's [timing allow passed flag](https://fetch.spec.whatwg.org/#concept-response-timing-allow-passed) is unset, then it returns `""` instead. This is probably slightly conservative since it would protect it if a redirect didn't pass the TAO check even if the final resource did, but similar concerns might apply to the destination of the redirect and there's value in being consistent with other TAO protected properties.

jeremyroman:
On further investigation, I think the current text in this PR already implies a TAO check in the following way:

Fetch says: 
> If response’s [timing allow passed flag](https://fetch.spec.whatwg.org/#concept-response-timing-allow-passed) is not set, then set timingInfo to a the result of [creating an opaque timing info](https://fetch.spec.whatwg.org/#create-an-opaque-timing-info) for timingInfo, set bodyInfo to a new [response body info](https://fetch.spec.whatwg.org/#response-body-info), and set cacheState to the empty string.

This is then used to populate the cache mode and thus the default delivery type, so "setup the resource timing entry" won't see the `"cache"` delivery type if the timing allow passed flag wasn't set (i.e., the response didn't pass a TAO check).

jeremyroman:
Status update:
* the question of the TAO is, I believe, resolved above
* re. TPAC discussion, user agents which restrict access to certain properties beyond the specified TAO checks can, of course, apply similar restrictions to this property if applicable; if explicit verbiage to permit this lands in the specification then I think it will generalize naturally
* slated to go to origin trial in Chrome 110 (including an extension for `"navigational-prefetch"`, which is otherwise difficult to detect)
* WPT coverage is landed as tentative ([resource-timing/delivery-type.tentative.any.js](https://github.com/web-platform-tests/wpt/blob/master/resource-timing/delivery-type.tentative.any.js))

I believe that covers the identified items. @yoavweiss what is the next step under the group's working mode?

yoavweiss:
^^ @achristensen07 @bdekoz

jeremyroman:
#369 seems to be general verbiage that I think covers any application it would have to this PR. I'd like to renew my request to know what the next step/milestone here is.

noamr:
This is LGTM, @yoavweiss do you want to take another look?

yoavweiss (approved):
LGTM

PR #350, opened 2022-09-29
Update index.html

caribouW3:
removed instability warning from sotd.

### DIFF
--- a/index.html
+++ b/index.html
@@ -79,14 +79,6 @@
       </p>
     </section>
     <section id="sotd">
-      <p class='warning'>
-        Implementers SHOULD be aware that this document is not stable.
-        Implementers who are not taking part in the discussions are likely to
-        find the specification changing out from under them in incompatible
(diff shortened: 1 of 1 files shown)


yoavweiss:
LGTM

yoavweiss (approved):

tunetheweb:
Didn’t this leave an empty section behind that’s not needed anymore now?

tunetheweb:
Ah never mind. I see automation caught this in #351 - nice!

caribouW3:
Originally I did not mean to merge before publication of the CR snapshot (on next tuesday) but anyway the ED still says it's a draft so it's OK. I'll also make the necessary changes on auto-publish next week.

yoavweiss:
Oh apologies! Should we revert?

caribouW3:
No harm done! The current ED has the same content anyway, it's not any less stable than what we're about to publish as CRS.

PR #361, opened 2022-11-14
Explainer for Content-type

abinpaul1:


Explainers/Content-Type.md
```diff
@@ -0,0 +1,150 @@
+# Content-Type in Resource Timing
+
+Proposal to add a new field `contentType` to PerformanceResourceTiming which holds a string corresponding to the content type of the fetched resource
```
yoavweiss, line 3: "corresponding to the content type header"?

Explainers/Content-Type.md
```diff
@@ -0,0 +1,150 @@
...
+const entry_list = performance.getEntriesByType("resource");
+console.log(entry_list[0].contentType);
+```
+
+
+## Content type Values
+
+The content type value would reflect the [Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) header on the response. It does NOT return the sniffed value.
+
+It would be an empty string if the [Cross-Origin-Resource-Sharing](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) [check](https://fetch.spec.whatwg.org/#concept-cors-check) fails
```
yoavweiss, line 39: Nit: "." at the end of the sentence?

yoavweiss (approved):
LGTM

mmocny:
Issue triage: closing, now that this work is "done" and we don't benefit from a dedicated explainer doc.

PR #366, opened 2022-12-18
Add interim response times

### LINKED ISSUE #345: Consider adding finalResponseHeaders{Start|End} times
Since we've added early hints, now TTFB in RUM, which maps to `responseStart` means something slightly different than it did before (though technically it represents the same thing).

For services that rely on final response timing, perhaps we should also expose when the final headers started arriving, and when they ended and the body started?

noamr:
Add `firstInterimResponseStart`: the first response with an informational (1XX) status.

Bug: #345
Depends on whatwg/fetch#1483

For now not adding headers-end and body-start as they're not planned to be implemented.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/pull/366.html" title="Last updated on May 8, 2023, 9:41 AM UTC (b82c99e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/366/95c8d0b...b82c99e.html" title="Last updated on May 8, 2023, 9:41 AM UTC (b82c99e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -367,7 +367,10 @@
               readonly attribute DOMHighResTimeStamp connectEnd;
               readonly attribute DOMHighResTimeStamp secureConnectionStart;
               readonly attribute DOMHighResTimeStamp requestStart;
+              readonly attribute DOMHighResTimeStamp firstInterimResponseStart;
               readonly attribute DOMHighResTimeStamp responseStart;
+              readonly attribute DOMHighResTimeStamp responseHeadersEnd;
+              readonly attribute DOMHighResTimeStamp responseBodyStart;
               readonly attribute DOMHighResTimeStamp responseEnd;
               readonly attribute unsigned long long  transferSize;
               readonly attribute unsigned long long  encodedBodySize;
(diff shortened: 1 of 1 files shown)


yoavweiss:
Seems reasonable once the Fetch bits land.

index.html
```diff
@@ -992,7 +1016,10 @@ <h3>
           {{PerformanceResourceTiming/connectStart}},
           {{PerformanceResourceTiming/connectEnd}},
           {{PerformanceResourceTiming/requestStart}},
+          {{PerformanceResourceTiming/firstInterimResponseStart}},
           {{PerformanceResourceTiming/responseStart}},
+          {{PerformanceResourceTiming/responseHeadersEnd}},
+          {{PerformanceResourceTiming/responseBodyStart}},
```
johnathan79717, line 1022: Thanks for making sure that the new timings require CORS. This is an essential requirement mentioned in the Chrome Security & Privacy intent review in order to avoid cross-origin leaks.

annevk:
Why only the first 103? What about a 109?

annevk:
Also, should we call this "interim" or "Informational"? Is there any precedent in the web platform?

noamr:
> Why only the first 103? What about a 109?

Because it's the only informational/interim header processed by the web platform.

noamr:
> Also, should we call this "interim" or "informational"? Is there any precedent in the web platform?

That was the proposal on the issue and it seemed ok at the time. I don't really mind renaming it to informational. @tunetheweb @yoavweiss any objections?

annevk:
> Because it's the only informational/interim header processed by the web platform.

Well, we handle all of them. We have some special code for 101 and 103, but it's not like we bail out when we see a 102.

tunetheweb:
@LPardue from the IETF HTTP WG had [some strong opinions on "informational"](https://github.com/w3c/resource-timing/issues/345#issuecomment-1234737040):

> I disagree with calling things informational responses. RFC 9110 doeant use that term, instead it defines interim responses, which use informational status codes. Terminology needs to be precise and consistent unless there is a good reason to diverge.

noamr:
> @LPardue from the IETF HTTP WG had [some strong opinions on "informational"](https://github.com/w3c/resource-timing/issues/345#issuecomment-1234737040):
> 
> > I disagree with calling things informational responses. RFC 9110 doeant use that term, instead it defines interim responses, which use informational status codes. Terminology needs to be precise and consistent unless there is a good reason to diverge.

Right, RFC9110 does use the term "interim". @annevk does that satisfy "precedent in the web platform"?

annevk:
It certainly does, thanks for digging that up.

noamr:
> > Because it's the only informational/interim header processed by the web platform.
> 
> Well, we handle all of them. We have some special code for 101 and 103, but it's not like we bail out when we see a 102. (Also the current Fetch PR is not restricted to a 103.)

By handle I mean actually do something meaningful with them. Perhaps this could be named `firstEarlyHintResponse` and be specific to the first 103?

annevk:
Right, but that means we should measure the first 1xx (if any), not the first 103. Which I tend to agree with.

noamr:
> > I think @LPardue (and @annevk ?)'s point is that it shouldn't be limited to 103 (even if it is for now) in case we want to use it for any other 1XX responses in future.
> 
> OK with me. So in conclusion I'll change the description & spec & implementation to also include 100, and we'll keep the `firstInterimResponseTime` name?

Done. Also removed the two unimplemented timings.

annevk:
It seems fine to expose less, but https://github.com/w3c/resource-timing/issues/345#issuecomment-1259852569 did have use cases for the others so that should still be tracked in some way I suppose?

Does this change also impact the request over at https://github.com/WebKit/standards-positions/issues/109?

noamr:
> It seems fine to expose less, but [#345 (comment)](https://github.com/w3c/resource-timing/issues/345#issuecomment-1259852569) did have use cases for the others so that should still be tracked in some way I suppose?


There is a use case for the others but decided to implement them gradually. Currently only `firstInterimResponseStart` is implemented. I can add all of them to the spec but I thought this would be a bit fictional. Perhaps we can leave the issue open for the other ones. 

> 
> Does this change also impact the request over at [WebKit/standards-positions#109](https://github.com/WebKit/standards-positions/issues/109)?

The standard position should consider that `firstInterimResponseStart` is the part of it that's currently implemented and the rest are in backlog for Chromium.

yoavweiss (approved):
LGTM

PR #385, opened 2024-03-25
Expose Content-Encoding to resource-timing

### LINKED ISSUE #381: Expose Content-Encoding to ResourceTiming
Hi,

Similar to request https://github.com/w3c/resource-timing/issues/203 to expose `Content-Type`, I would like to request exposing the `Content-Encoding` of each resource to ResourceTiming.

As we're starting to see experimentation and deployments of new content encodings such as Zstandard (`zstd`) and [compression dictionary transports](https://github.com/WICG/compression-dictionary-transport) (`zstd-d` and `br-d`), we are moving toward content being delivered from a large set of possible encodings, even to the same client on different page loads or (sub)requests to the same domain.

When the content encoded was a small set: (none), gzip and brotli, one could often infer the encoding depending on the encoded/decoded body sizes, though that generally only works if one "owns" the content (has visibility into what the size would be for each encoding type).

Having an explicit `.contentEncoding` would help with some use-cases I can think of:

* As origins and CDNs want to experiment with new Content Encodings, they could utilize phased rollouts of the above technologies and be able to segment (from RUM) the performance of clients utilizing those new technologies.
    * For example, this can help with understanding real-world uptake %, how the performance behaves for those new technologies, and debugging them
* Once fully deployed, origins and CDNs could segment browsing sessions (or A/B test) with those technologies to provide value confirmation and/or debugging via RUM.
    * For example, one could verify with RUM (along serverside logs) how often dictionaries are being used.
* Third-party RUM vendors could segment resource fetches by their encoding with certainty (rather than guessing based on file sizes)
    * For example, RUM vendors could offer insights and suggestions to enable more advanced encodings and/or dictionaries

CC @pmeenan @horo-t

Jxck:
add `contentEncoding` to Resource Timing.
closed #381.

### DIFF
--- a/index.html
+++ b/index.html
@@ -376,6 +376,7 @@
               readonly attribute unsigned short responseStatus;
               readonly attribute RenderBlockingStatusType renderBlockingStatus;
               readonly attribute DOMString contentType;
+              readonly attribute DOMString contentEncoding;
               [Default] object toJSON();
           };
         </pre>
@@ -708,6 +709,11 @@
           The <dfn>contentType</dfn> getter steps are to return <a>this</a>'s
           <a data-for="PerformanceResourceTiming">resource info</a>'s
           [=response body info/content type=].
+        </p>
+        <p data-dfn-for="PerformanceResourceTiming">
+          The <dfn>contentEncoding</dfn> getter steps are to return <a>this</a>'s
+          <a data-for="PerformanceResourceTiming">resource info</a>'s
+          [=response body info/content encoding=].
         </p>
         <p data-dfn-for="PerformanceResourceTiming">
           The <dfn>renderBlockingStatus</dfn> getter steps are to return


index.html
```diff
@@ -709,6 +710,11 @@ <h3>
           <a data-for="PerformanceResourceTiming">resource info</a>'s
           [=response body info/content type=].
         </p>
+        <p data-dfn-for="PerformanceResourceTiming">
+          The <dfn>contentEncoding</dfn> getter steps are to return <a>this</a>'s
+          <a data-for="PerformanceResourceTiming">resource info</a>'s
+          [=response body info/content encoding=].
```
horo-t, line 716: The Fetch spec also needs to be updated.
The Fetch spec side PR will be similar to https://github.com/whatwg/fetch/pull/1481.

Jxck: Filed https://github.com/whatwg/fetch/pull/1742

Soonerseye (approved):

PR #387, opened 2024-04-26
Correct clarification for data: URIs

### LINKED ISSUE #386: Should data: urls be reported in resource timing?
Context: https://issues.chromium.org/issues/333831248

The "Resources included" section of the spec says (in a non-normative list) that

> If an HTML IMG element has a [data: URI](https://tools.ietf.org/html/rfc2397) as its source [[RFC2397](https://w3c.github.io/resource-timing/#bib-rfc2397)], then this resource will not be included as a [PerformanceResourceTiming](https://w3c.github.io/resource-timing/#dom-performanceresourcetiming) object in the [Performance Timeline](https://www.w3.org/TR/performance-timeline/#performance-timeline). By definition [data: URI](https://tools.ietf.org/html/rfc2397) contains embedded data and does not require a [fetch](https://fetch.spec.whatwg.org/#concept-fetch).

This text has been present in the spec as far back as I can trace its history -- back to 2011, at least. However, since 2013, data: urls *have* been handled by fetch (see https://fetch.spec.whatwg.org/#data-urls), and there's no normative spec text in resource timing that I can see that prohibits them.

There's no WPT validating either behaviour, and I think that user agents may be in disagreement on whether to include them or not.

Should we remove that note? Or add a check in the spec to exclude data: urls? (and possibly blob urls, and anything else that shouldn't require a network fetch)

noamr:
Closes #386

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/pull/387.html" title="Last updated on Apr 26, 2024, 3:04 PM UTC (0226ea5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/resource-timing/387/c75a9f5...0226ea5.html" title="Last updated on Apr 26, 2024, 3:04 PM UTC (0226ea5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -322,11 +322,10 @@
           <li>If an HTML <code>IMG</code> element has a <code><a href=
           "https://tools.ietf.org/html/rfc2397">data: URI</a></code> as its
           source [[RFC2397]], then this resource will not be included as a <a>
-            PerformanceResourceTiming</a> object in the <a data-cite=
-            "PERFORMANCE-TIMELINE-2#performance-timeline">Performance
-            Timeline</a>. By definition <code><a href=
-            "https://tools.ietf.org/html/rfc2397">data: URI</a></code> contains
-            embedded data and does not require a [=fetch=].
+          PerformanceResourceTiming</a> object in the <a data-cite=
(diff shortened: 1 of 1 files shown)


clelland (approved):
Thanks!

PR #408, opened 2024-11-04
Add finalResponseHeadersStart which is always the 2xx/4xx/5xx response.

### LINKED ISSUE #345: Consider adding finalResponseHeaders{Start|End} times

noamr:
responseStart now returns `firstInterimResponseStart` as before, but falls back to `finalResponseHeadersStart` if there is no interim response.

Closes #345

### DIFF
--- a/index.html
+++ b/index.html
@@ -366,6 +366,7 @@
               readonly attribute DOMHighResTimeStamp connectEnd;
               readonly attribute DOMHighResTimeStamp secureConnectionStart;
               readonly attribute DOMHighResTimeStamp requestStart;
+              readonly attribute DOMHighResTimeStamp finalResponseHeadersStart;
               readonly attribute DOMHighResTimeStamp firstInterimResponseStart;
               readonly attribute DOMHighResTimeStamp responseStart;
               readonly attribute DOMHighResTimeStamp responseEnd;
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -636,12 +637,17 @@ <h3>
           global object</a> for <a>this</a>. See [=/HTTP fetch=] for more info.
         </p>
         <p data-dfn-for="PerformanceResourceTiming">
-          The <dfn>responseStart</dfn> getter steps are to <a>convert fetch
+          The <dfn>finalResponseHeadersStart</dfn> getter steps are to <a>convert fetch
```
yoavweiss, line 640: I *thought* we settled on `finalResponseHeaderStart` (singular rather than plural). Did I get it wrong?

noamr: Nope. Look at #345. It's when the headers (plural) arrive.

tunetheweb: I don't remember being clear one way or another. I documented `finalResponseHeadersStart` but without too much thought to be honest. I don't feel strongly one way or the other.

Is there any precedent either way?

noamr: It's "header" in the HTTP spec but [`Headers`](https://fetch.spec.whatwg.org/#headers-class) in `fetch`, especially when web-exposed (`fetch(url).headers`). I think we're closer to `fetch` here and should align with the web-expose notion of `Headers`.

LPardue: I'd prefer headers

yoavweiss: As @noamr pointed out elsewhere, the [http terminology](https://www.rfc-editor.org/rfc/rfc9110.html#name-header-fields) is (coloquially) "headers" and Fetch has a [Headers](https://fetch.spec.whatwg.org/#headers-class) class, so let's align with that

yoavweiss (approved):
LGTM

tunetheweb:
We probably should update the diagram for this: https://w3c.github.io/resource-timing/#attribute-descriptions

Though it's quite old anyway and has diverged from the Navigation Timing one: https://w3c.github.io/navigation-timing/#processing-model

tunetheweb:
Raised these:
- Resource Timing - https://github.com/w3c/resource-timing/pull/410
- Navigation Timing - https://github.com/w3c/navigation-timing/pull/205

PR #418, opened 2025-06-26
Edit the list of possible initiatorType values.

guohuideng2024:
Update the list.

### DIFF
--- a/index.html
+++ b/index.html
@@ -439,6 +439,11 @@
               request=];
             </li>
             <li>
+              <code>"body"</code>, if the request is a result of processing
+              the [^body^] element's <code>background</code> attribute that's
+              already obsolete.
+            </li>
+            <li>
               <code>"css"</code>, if the request is a result of processing a
               CSS <a data-cite="css-values-4" data-xref-type=
               "css-function">url()</a> directive such as <code>@import
@@ -487,8 +492,8 @@
               an [^input^] element of [^input/type^] [^input/type/image^].
             </li>
             <li>
-              <code>"a"</code>, if the request is the result of processing an
-              [^a^] element's [^a/download^] or [^a/ping^].
+              <code>"ping"</code>, if the request is the result of processing an
+              [^a^] element's [^a/ping^].
             </li>
             <li>
               <code>"iframe"</code>, if the request is the result of processing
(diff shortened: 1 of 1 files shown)


guohuideng2024:
I made this edit mostly based on https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/initiatorType, Chromium and WPT test.

Decision of the values modified/added here is based on
https://github.com/w3c/resource-timing/issues/364#issuecomment-3006033584

Below is the list of things that are not obvious. And I need some guidance on some of them.

1)  For the "body" value, it's used for the "background" attribute in "body" element, which is obsolete now. That's why there is no reference link to the current html standard.

**I am not sure if the current text is appropriate, and I am not sure if I should include the "body" value in this list because this is obsolete now.**

2)  Looks like the "ping" replaced "a" and it's for the "ping" attribute only.

3) The link to "Early hints" is broken. I cannot find an exported reference to "early hint" on
https://respec.org/xref/
and I see the "fetch" standard is not refereeing it neither. (see the third note on 
https://fetch.spec.whatwg.org/#fetching )

**So I plan to submit a PR that adds a `class="export"` to the "Early hints" dfn in html spec. Is it a right move?**

Thanks.

guohuideng2024:
@noamr  : You mentioned `font` is used by IFT, I think I should add `font` here? 

From https://www.w3.org/TR/IFT/, I see the following text:

"This document was produced by the [Web Fonts Working Group](https://www.w3.org/groups/wg/webfonts) as a Working Draft using the [Recommendation track](https://www.w3.org/policies/process/20231103/#recs-and-notes). This document is intended to become a W3C Recommendation."

So it looks like it's not yet adopted and implemented. (No wonder I cannot find it in Chromium and WPT). 
Should I simply say something like: `font` value --> downloaded using IFT technology?

Thanks.

index.html
```diff
@@ -439,10 +451,25 @@ <h3>
...
+            <li>
+              <p>
+                <code>"css"</code>, if the request is a result of processing a
+                CSS <a data-cite="css-values-4" data-xref-type=
+                "css-function">url()</a> directive such as <code>@import
+                url()</code> or <code>background: url()</code>; [[CSS-VALUES]]
+              </p>
+
+              <p>
+                Another example is, the request for a font resource specified
```
noamr, line 467: "Note:" instead of "Another example is,"

PR #426, opened 2026-02-17
Remove index.html from gh-pages

yoavweiss:


mmocny:
Hi @yoavweiss Does this still need review and need landing?

I'm a bit confused, since I think the workflow is to build from main and generate an artifact into the gh-pages branch automatically.  Is that not meant to be checked in?

yoavweiss:
Probably not..

